PR #1, opened 2019-10-25
Implement sever side rendering for SEO

jorilindell:


client/.env
```diff
@@ -0,0 +1,2 @@
+REACT_APP_GRAPHQL_BASE_URL=http://localhost:4000/proxy/graphql
```
badrange: These ones could go to `.env.development` (and that move can of course wait until we have a "production" environment)

badrange:
Looks fine to me. I have so little experience with SSR that my opinion carries little weight. 

One thing that would be nice is more instructions on how to do debugging, especially of the SSR server.

client/config/env.js
```diff
@@ -0,0 +1,72 @@
+const fs = require("fs");
```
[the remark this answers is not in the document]
jorilindell: Yeah, these webpack related files are more or less same that create-react-app uses under the hood, but there is now way to re-use them without ejecting. These are for SSR to build the project

client/.env
jorilindell: I think that I'll wait to the point when we are talking about deployment

PR #2, opened 2019-10-25
Basic structure of GraphQL proxy

jorilindell:


graphql-proxy/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "eslint-plugin-jsx-a11y": "^6.2.3",
+    "eslint-plugin-prettier": "^3.1.1",
+    "eslint-plugin-react": "^7.16.0",
+    "eslint-plugin-react-hooks": "^1.7",
+    "eslint-plugin-simple-import-sort": "^4.0.0",
+    "prettier": "^1.18.2"
+  },
+  "jest": {
+    "setupFilesAfterEnv": [
+      "./test/jest.setup.js"
```
jimppa: This file does not exist, causing `yarn test` to fail.

badrange:
Looks good to me; and it would be great with instructions for debugging (even configurations in .vscode/launch.json committed to repo) here too.

graphql-proxy/src/schema/resolvers.ts
```diff
@@ -0,0 +1,7 @@
+import { merge } from "lodash";
```
jimppa: I think this imports all of lodash, import instead just the merge function from `lodash/merge`

graphql-proxy/package.json
jorilindell: FIxed, setup file was on wrong folder

graphql-proxy/src/schema/resolvers.ts
jorilindell: Changed

graphql-proxy/src/schema/schema.ts
```diff
@@ -0,0 +1,6 @@
+import dummy from "./dummy/typeDefs";
```
jimppa: Does this file have any purpose? It's identical to `typeDefs.ts`, but this file is not used.

PR #8, opened 2019-11-19
Home page hero component

jorilindell:


client/src/common/components/hero/hero.module.scss
```diff
@@ -0,0 +1,62 @@
...
+.heroWrapper {
+  position: relative;
+  width: 100%;
+  height: 42.6875rem;
+  background-image: url("../../../assets/icons/svg/hero-background.svg");
+  background-position: center;
+  background-size: cover;
+  background-repeat: no-repeat;
+
+  @include respond-below(sm) {
```
jimppa: Not sure if this matters much here, but wouldn't it make more sense to write mobile styles first and override on bigger screens?

client/src/domain/home/search.module.scss
```diff
@@ -2,7 +2,7 @@
 @import "layout";
 
 .searchContainer {
-  margin-top: 5rem;
+  margin-top: -5.5rem;
```
jimppa: Would it be possible to make this adjustment in the component that uses `<Search />` instead? Or is this needed for something that this component caused?

PR #11, opened 2019-12-02
TH-84: Event page hero

jorilindell:
https://helsinkisolutionoffice.atlassian.net/browse/TH-84

Only the hero component is implemented in scope of this PR. The task is huge so it doesn't make sense to implement everything in one PR


client/src/util/createClient.tsx
```diff
@@ -0,0 +1,27 @@
+import { InMemoryCache } from "apollo-cache-inmemory";
+import ApolloClient from "apollo-client";
+import { ApolloLink } from "apollo-link";
+import { BatchHttpLink } from "apollo-link-batch-http";
+import { createPersistedQueryLink } from "apollo-link-persisted-queries";
+import get from "lodash/get";
+
+export default function createClient(uri: string) {
```
jimppa: There is nothing in this file indicating that this is a graphql-client, maybe rename file and function to createGraphqlClient?

jorilindell: Good point. Renamed

graphql-proxy/src/utils/toCamelCase.ts
```diff
@@ -0,0 +1,9 @@
+export default (snakecase: string) => {
```
badrange: What about adding some comments explaining what this function does? A couple of before-after examples would save a lot of brainpower.

graphql-proxy/src/utils/objectToCamelCase.ts
```diff
@@ -0,0 +1,27 @@
+import memoize from "lodash/memoize";
+
+import toCamelCase from "./toCamelCase";
+
+const memoizedCamelCase = memoize(toCamelCase);
+
+const objectToCamelCase = value => {
```
badrange: An explanation of what and why could be helpful for a future developer joining the project

client/src/util/getLocalisedString.ts
```diff
@@ -0,0 +1,22 @@
...
+import { Language } from "../types";
+
+export default (obj: LocalizedObject, language: Language) => {
+  const languages = [
+    language,
+    ...Object.values(SUPPORT_LANGUAGES).filter(item => item !== language)
+  ];
+  let str = "";
+
+  forEach(languages, lng => {
```
jimppa: [`Array.prototype.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) does exactly what you are doing here

jimppa: Also, a unit-test would be helpful documentation for this module

client/src/common/components/keyword/keyword.module.scss
```diff
@@ -0,0 +1,25 @@
...
+  background-color: $hel-summer-light-50;
+  color: $hel-black;
+  margin: 0.1875rem 0.375rem 0.1875rem 0;
+  display: inline-flex;
+  padding: 0.125rem 0.5rem;
+  align-items: center;
+  font-size: 0.875rem;
+  font-weight: 500;
+
+  @include respond_below(sm) {
```
jimppa: Would prefer mobile first in the future

client/src/icons/AngleRightIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+type Props = { className?: string };
+
+export default ({ className = "" }: Props) => (
```
jimppa: If you are using CRA, you can add `className` to svg-icons just fine. So I would have this module just as svg-module instead, no need for tsx.

client/src/domain/event/EventHero.tsx
```diff
@@ -0,0 +1,285 @@
...
+              )}
+              <div className={classNames(styles.date, styles.desktopOnly)}>
+                {getDateStr(startTime, endTime)}
+              </div>
+              <div className={styles.title}>
+                {getLocalisedString(name, locale)}
+              </div>
+              <div
+                className={styles.description}
+                dangerouslySetInnerHTML={{
```
jimppa: Why does this need to be set dangerously?

client/src/domain/event/EventHero.tsx
```diff
@@ -0,0 +1,285 @@
...
+                    keywords.map(keyword => {
+                      return (
+                        <Keyword
+                          key={keyword.id}
+                          keyword={getLocalisedString(keyword.name, locale)}
+                        />
+                      );
+                    })}
+                  {today && !thisWeek && (
+                    <Keyword color="lightEngel50" keyword="Today" />
```
jimppa: Should this keyword be translated?

client/src/domain/event/EventHero.tsx
```diff
@@ -0,0 +1,285 @@
...
+    );
+    const streetAddress = getLocalisedString(
+      (location && location.streetAddress) || {},
+      locale
+    );
+
+    return [locationName, streetAddress, addressLocality].join(", ");
+  };
+
+  const getDateStr = (start: string, end: string | null | undefined) => {
```
jimppa: Would it make sense to have this function as an utility function in some separate file?

client/src/domain/event/EventHero.tsx
jorilindell: I forgot to change this. It used to have data that contains html. But not needed any more

PR #12, opened 2019-12-05
TH-153 | Event location map

jorilindell:
https://helsinkisolutionoffice.atlassian.net/browse/TH-153


client/src/domain/event/EventLocation.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const name = eventData.linkedEventsEventDetails.name;
+
+  return (
+    <div className={styles.eventLocationContainer}>
+      <div className={styles.titleWrapper}>
+        <div className={styles.title}>
+          <LocationIcon />
+          <h3>{t("event.location.title")}</h3>
+        </div>
+        <a className={styles.mapLink} href={getGoogleLink()} target="__blank">
```
jimppa: `__blank` should be just `_blank` with a single underscore. Remember to add `rel="noopener noreferrer"`

client/src/domain/event/EventLocation.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+interface Props {
+  eventData: EventDetailsQuery;
+}
+
+const EventLocation: React.FC<Props> = ({ eventData }) => {
+  const { t } = useTranslation();
+  const locale = getLocale();
+
+  const getGoogleLink = () => {
```
jimppa: Could these `getX`-functions be declared outside this function? Now the `getX`-functions are recreated on every render.

jorilindell: Good point. Already fixed that on PR13

client/src/common/components/map/Map.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+import L, { Icon } from "leaflet";
+import React from "react";
+import { Map as LeafletMap, Marker, TileLayer } from "react-leaflet";
+
+import markerIconUrl from "../../../assets/images/png/map-marker.png";
+import styles from "./map.module.scss";
+
+// Allow map scroll only inside Helsinki
+const maxBounds: [number, number][] = [[60.33, 25.33], [60.1, 24.73]];
```
jimppa: Would it be possible to use some more readable type for these lat/lng tuples? I have a hard time knowing which value is lat or lng.

PR #13, opened 2019-12-09
TH-155 | Event page fields

jorilindell:
https://helsinkisolutionoffice.atlassian.net/browse/TH-155


client/src/domain/event/EventContent.tsx
```diff
@@ -0,0 +1,50 @@
...
+      </div>
+      <div className={styles.descriptionColumn}>
+        {description && (
+          <>
+            <h3 className={styles.descriptionTitle}>
+              {t("event.description.title")}
+            </h3>
+            <div
+              className={styles.description}
+              dangerouslySetInnerHTML={{
```
jimppa: Since this is set dangerously, `description` should be sanitized.

client/src/domain/event/EventInfo.tsx
```diff
@@ -0,0 +1,238 @@
...
+            <div className={styles.desktopOnly}>{streetAddress}</div>
+          )}
+          {district && <div className={styles.desktopOnly}>{district}</div>}
+          {addressLocality && (
+            <div className={styles.desktopOnly}>{addressLocality}</div>
+          )}
+          <a
+            className={styles.link}
+            href={getGoogleLink(eventData, locale)}
+            target="__blank"
```
jimppa: `_blank`

client/src/domain/event/EventHero.tsx
```diff
@@ -55,47 +54,31 @@ const EventHero: React.FC<Props> = ({
       locale
     );
 
-    return [locationName, streetAddress, addressLocality].join(", ");
+    return [locationName, streetAddress, addressLocality]
```
jimppa: I have seen this getLocationStr pattern so many times that I think it should be an own component.

PR #14, opened 2019-12-11
TH-160 | Event search page without filtering

jorilindell:
https://helsinkisolutionoffice.atlassian.net/browse/TH-160


client/src/domain/event/EventInfo.tsx
```diff
@@ -157,7 +161,12 @@ const EventInfo: React.FC<Props> = ({ eventData }) => {
             {email && <div>{email}</div>}
             {telephone && <div>{telephone}</div>}
             {infoUrl && (
-              <a className={styles.link} href={infoUrl || ""} target="__blank">
+              <a
+                className={styles.link}
+                href={infoUrl || ""}
```
jimppa: infoUrl is guaranteed to be something, it is checked already before rendering this tag

client/src/domain/event/EventUtils.ts
```diff
@@ -1,54 +1,54 @@
...
-  const location = eventData.eventDetails.location;
-  const district =
-    location &&
-    location.divisions.find(
-      division => ["district", "neighborhood"].indexOf(division.type) !== -1
-    );
+  const location = event.location;
+  const divisions = (location && location.divisions) || [];
+  const district = divisions.find(
+    division => ["district", "neighborhood"].indexOf(division.type) !== -1
```
jimppa: I think `Array.prototype.includes` would be more readable than `indexOf`

PR #15, opened 2019-12-12
TH-154 | Display similar events on event page

jorilindell:
https://helsinkisolutionoffice.atlassian.net/browse/TH-154


client/src/domain/event/EventPageContainer.tsx
```diff
@@ -26,6 +28,7 @@ const EventPageContainer: React.FC<
               <EventHero eventData={eventData} />
               <Container>
                 <EventContent eventData={eventData} />
+                {isClient && <SimilarEvents eventData={eventData} />}
```
jimppa: Any reason for rendering this only in client?

jorilindell: Fetching list of event takes over 1.5 seconds so it makes initial load quite slow

PR #30, opened 2020-01-15
TH-191 | Use existing icons from HDS

Jokinen:
These changes will replace icons in our repo with ones that can be found from the HDS.

* LanguageIcon
* SmileIcon
* food
* menu
* influence → fill
* nature → tree
* search
* smile

These two are a special case in that they are used in `scss` files and require more refactoring

angle-right: X (ignored for now; hard to refactor, added comments)

area → location: X

The exception here is `angle-right` icon. It's used through a `.scss` file to amend a css class used by `react-datepicker`. I couldn't easily find a way to accomplish the replacement of this icon in another way so I moved it into a special folder to communicate its status.

https://helsinkisolutionoffice.atlassian.net/browse/TH-191?atlOrigin=eyJpIjoiNWIwYmJhMTlhNzIzNGRhM2E0YWNiOTVmMDYyMWQ4MGUiLCJwIjoiaiJ9


Jokinen:
Rebased on current `develop` to resolve merge conflicts.

PR #37, opened 2020-01-17
Skip fetching event list data on SSR to make initial load faster

jorilindell:


Jokinen:
Makes a pretty big difference 👍 Nice catch.

PR #39, opened 2020-01-17
TH-193 | Make dropdowns navigable with keyboard

Jokinen:
Allows user to navigate dropdowns with keyboard.

While I am writing this PR I made the recollection that this keyboard navigation work slightly differently to that of the search field. I'll change it on Mon if it's an issue.

Differences:  
* Scrolling "UP" from first item will set index to `-1`
* Scrolling "DOWN" from last item will set index to `-1`
* By default the "cursor" is in position `-1`
* I just copied these from Google

I didn't do that many accessibility considerations. When the user navigates to a hidden item on a list, it's scrolled into view. No other attributes or anything like that.

I needed to edit the css rules for checkbox to allow its padding to be extended. This allows the focus state to paint it in a more visually appealing way at the same time ensuring consistency between clicks and keyboard. It shouldn't probably be an issue because I don't think the checkbox was used anywhere else.

https://helsinkisolutionoffice.atlassian.net/browse/TH-193?atlOrigin=eyJpIjoiZjAxNTYxNjhjNmRmNDQ1MTkxYmZhNDU4MGIzMDIwZmMiLCJwIjoiaiJ9


Jokinen:
Made improvements suggested by @jorilindell and reabsed on top of current `develop`.

Jokinen:
Rebased on current develop.

PR #43, opened 2020-01-20
TH-219 | Button to clear event search filters

jorilindell:
https://helsinkisolutionoffice.atlassian.net/browse/TH-219

### DIFF
--- a/src/domain/eventSearch/SearchResultList.tsx
+++ b/src/domain/eventSearch/SearchResultList.tsx
@@ -102,6 +102,26 @@ const SearchResultList: React.FC<Props> = ({
     window.scrollTo(0, 0);
   };
 
+  const clearFilters = () => {
+    const search = getSearchQuery({
+      categories: [],
+      dateTypes: [],
+      districts: [],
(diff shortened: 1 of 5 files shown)


src/domain/eventSearch/SearchResultList.tsx
```diff
@@ -102,6 +102,26 @@ const SearchResultList: React.FC<Props> = ({
...
+      places: [],
+      publisher: null,
+      search: "",
+      startDate: null,
+      targets: []
+    });
+
+    push({ pathname: `/${locale}/events`, search });
+    // Scroll to top when changing filters. Ignore this on SSR becasue window doesn't exist
+    window.scrollTo(0, 0);
```
Jokinen: The comment here describes that the window line should be ignored when using SSR, but there doesn't seem to be any explicit condition that does that. As far as I understand, with this code, the `scrollTo` action would not be ran on client side though.

jorilindell: So true, I'll remove scrollTo

PR #52, opened 2020-01-23
Fix language init and support default browser language

Jokinen:
Previously translations were not actually passed down from server to client which resulted in a "flash of untranslated content". This wouldn't happen when you used an `url` that had a language parameter so it was easy to miss.

I made the client respect the language passed down from the server. This only works when you navigate into the root of the app. Either the cache or `i18next` set a pretty sticky language choice so this will reliably work only when accessing the app for the first time (or using Incognito/etc).

Issues: The build uses the translation files from within `src`. This doesn't seem correct to me, but it was the approach the official example used for SSR. Should this be resolved?


Jokinen:
So, yeah, the translation files being set up this way is a problem. We should make sure they get included in the build files and use them from there.

This should not be merged before this issue is resolved.

Jokinen:
I removed `i18next-node-fs-backend` from use, because to my understanding we do not benefit from having it because we transpile our server code.

Jokinen:
The build seems to be finding the translation files now 👍

PR #54, opened 2020-01-23
Feature/language change on server

jorilindell:
#52 fixed problem that translated strings were not passed from server to client. After those changes the ui is working only in Finnish.

Problem is that correct language is not set on the server, but the default language is used. Also the   initialI18nStore includes only Finnish translations.

This PR fixes those issues


Jokinen:
I tested this out a bit and for me it doesn't seem to work expectedly. I use English as my browser language, but when I access the build from incognito mode, I am shown the Finnish version.

On the old branch, the initial languages I received were: `fi` and `en`. I wonder if it's because `fi` is fallback and my browser was using `en`? Maybe you only got `fi` because your browser is in Finnish and the fallback language is also Finnish?

The purpose of this behaviour may be to decrease the amount of data that gets passed around down when you have multiple language versions--it shouldn't matter that much for this app though because we are only supporting three so including all three sounds like a good solution.

PR #81, opened 2020-03-13
TH-408 | Implements script to generate sitemaps in xml format

jorilindell:
## Description :sparkles:
1. Generates sitemap index in xml format that lists all sitemaps
2. Generates sitemaps (one per language) in xml format that lists urls for all the events in selected language
3. Generates sitemaps (one per language) in xml format that lists urls for all the collections in selected language

Jira ticket: https://helsinkisolutionoffice.atlassian.net/browse/TH-408

## Testing :alembic:
Feature can be tested by running `yarn update-sitemap`
Script should generate following files:

- public/sitemap.xml
- public/sitemap_collections_en.xml
- public/sitemap_collections_fi.xml
- public/sitemap_collections_sv.xml
- public/sitemap_events_en.xml
- public/sitemap_events_fi.xml
- public/sitemap_events_sv.xml


scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+
+/**
+ * Generate sitemap in xml format for events in all languages
+ */
+const generateEventSitemaps = async () => {
+  await Promise.all(
+    languages.map(async lang => {
+      try {
+        const events = await getEvents(lang);
+        // eslint-disable-next-line no-console
```
Nizar-Rahme: No need for this disable rule here

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+        await generateEventSitemap(events, lang);
```
Nizar-Rahme: Why do you need `await` here?

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+      } catch (err) {
+        throw err;
```
Nizar-Rahme: So catching the error here is redundant if you're going to throw it

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+  const collections = await getCollections();
+  await Promise.all(
+    languages.map(async lang => {
+      try {
+        await generateCollectionSitemap(collections, lang);
+      } catch (err) {
+        throw err;
+      }
+    })
+  );
```
Nizar-Rahme:
```suggestion
 Promise.all(languages.map(lang => await generateCollectionSitemap(collections, lang)));
```
I have same comments as above; I think this would do the trick.

scripts/update-sitemap.ts
Nizar-Rahme: Why do you need `await` here?
The callback you provided for `map` is returning `undefined` so you'll end up with an array of `undefined` passed to `Promise.all`

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+ * Write object to a xml file
+ * @param {string} path
+ * @param {object} data
+ */
+const writeXMLFile = async (path: string, data: object) => {
+  const options = { compact: true, ignoreComment: true, spaces: 4 };
+  const xml = convert.js2xml(data, options);
+
+  await writeFile(path, xml, "utf8");
+};
```
Nizar-Rahme:
```suggestion
const writeXMLFile = (path: string, data: object) => {
  const options = { compact: true, ignoreComment: true, spaces: 4 };
  const xml = convert.js2xml(data, options);

  return writeFile(path, xml, "utf8");
};
```

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+        ...events.map(event => ({
+          lastmod: formatDate(event.last_modified_time),
+          loc: `${host}/${language}/event/${event.id}`
+        }))
+      ]
+    }
+  };
+
+  await writeXMLFile(`${pathToSitemaps}/sitemap_events_${language}.xml`, data);
+};
```
Nizar-Rahme:
```suggestion
const generateEventSitemap = (
  events: SimpleEvent[],
  language: string
) => {
  const data = {
    _declaration: { _attributes: { encoding: "utf-8", version: "1.0" } },
    urlset: {
      _attributes: { xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9" },
      url: [
        ...events.map(event => ({
          lastmod: formatDate(event.last_modified_time),
          loc: `${host}/${language}/event/${event.id}`
        }))
      ]
    }
  };

  return writeXMLFile(`${pathToSitemaps}/sitemap_events_${language}.xml`, data);
};
```

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+        }))
+      ]
+    }
+  };
+
+  await writeXMLFile(
+    `${pathToSitemaps}/sitemap_collections_${language}.xml`,
+    data
+  );
+};
```
Nizar-Rahme:
```suggestion
const generateCollectionSitemap = (
  events: SimpleCollection[],
  language: string
) => {
  const data = {
    _declaration: { _attributes: { encoding: "utf-8", version: "1.0" } },
    urlset: {
      _attributes: { xmlns: "http://www.sitemaps.org/schemas/sitemap/0.9" },
      url: [
        ...events.map(event => ({
          lastmod: formatDate(event.last_published_at),
          loc: `${host}/${language}/event/${event.id}`
        }))
      ]
    }
  };

  return writeXMLFile(
    `${pathToSitemaps}/sitemap_collections_${language}.xml`,
    data
  );
};
```

scripts/update-sitemap.ts
```diff
@@ -0,0 +1,240 @@
...
+          loc: `${host}/sitemap_events_${language}.xml`
+        })),
+        ...languages.map(language => ({
+          loc: `${host}/sitemap_collections_${language}.xml`
+        }))
+      ]
+    }
+  };
+
+  await writeXMLFile(`${pathToSitemaps}/sitemap.xml`, data);
```
Nizar-Rahme: Same here!

scripts/update-sitemap.ts
```diff
@@ -189,14 +184,8 @@ const generateCollectionSitemap = async (
...
-  await Promise.all(
-    languages.map(async lang => {
-      try {
-        await generateCollectionSitemap(collections, lang);
-      } catch (err) {
-        throw err;
-      }
-    })
+  return Promise.all(
+    languages.map(async lang => generateCollectionSitemap(collections, lang))
```
Nizar-Rahme: `async` can be removed, otherwise, LGTM!

PR #93, opened 2020-06-02
TH-618 | Select all checkbox on multi-select dropdown

jorilindell:
## Description :sparkles:
1. Use Checkbox component from hds-react
2. Fix hero image position
3. Get single keyword from apollo cache if possible to improve performance
4.Fix collection title sizes
5. Implement select all checkbox to multi-select dropdown component


src/common/components/multiSelectDropdown/MultiSelectDropdown.tsx
```diff
@@ -176,7 +190,11 @@ const Dropdown: React.FC<Props> = ({
 
   const handleValueChange = (event: React.ChangeEvent<HTMLInputElement>) => {
     const val = event.target.value;
-    toggleOption(val);
+    if (val === SELECT_ALL) {
+      handleClear();
```
Sartonon, line 194: Whats does select all do? seems strange that it clears the array

jorilindell: Logic of the "select all" checkbox is described in https://helsinkisolutionoffice.atlassian.net/browse/TH-618 (I agree, it's little bit strange)

Maybe naming is little bit confusing here.

PR #96, opened 2020-08-19
TH-745 | Clear multi select dropdown input values

jorilindell:
## Description
- Clear multi select dropdown input values when clearing search filters
- Add getEventFilters helper function to get filter variables

## Closes
TH-745


src/domain/eventSearch/utils.ts
```diff
@@ -219,3 +224,72 @@ export const getNextPage = (
...
+    if (!isEmpty(filter) || isNumber(filter) || typeof filter === 'boolean') {
+      if (isArray(filter)) {
+        const items: Array<string | number> = [];
+
+        forEach(filter, (item: string | number) => {
+          items.push(encodeURIComponent(item));
+        });
+
+        query.push(`${key}=${items.join(',')}`);
+      } else if (filter != null) {
```
Sartonon, line 288: Maybe lodash isNil would be easier to understand than != null

Sartonon (approved):
Looking good. Added one comment, just a thought

PR #100, opened 2020-08-24
TH-648 | Add checkbox to show only evening events

jorilindell:
## Description
Add checkbox to filter evening events

## Issues
TH-648


Sartonon (approved):
Good!

PR #102, opened 2020-08-25
About page

jorilindell:
## Description
Implements about page

## Closes
**[TH-752](https://helsinkisolutionoffice.atlassian.net/browse/TH-752)**

### DIFF
--- a/src/common/translation/i18n/en.json
+++ b/src/common/translation/i18n/en.json
@@ -1,4 +1,7 @@
 {
+  "about": {
+    "title": "About the service"
+  },
   "appName": "Events",
   "commons": {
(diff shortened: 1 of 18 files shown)


Sartonon (approved):
Approved, but what is the reasoning to have own page component for all the translations and not having them in localisation files? Just wondering :)

jorilindell:
> Approved, but what is the reasoning to have own page component for all the translations and not having them in localisation files? Just wondering :)

At some point the content of the about page is going to be moved to CMS. So we have to rewrite this page anyway. Also the structure of each language is little bit different so I didn't want to implement too much logic to this simple page

PR #104, opened 2020-09-07
TH-770 | Sitemap improvements

jorilindell:
## Description
Update sitemap (if UPDATE_SITEMAP is true) when express server starts
Update sitemap everyday at 5:00 A.M.
Update robots.txt.  
Add static pages to sitemap
Fetch events parallel to improve performance

## Closes
[TH-770](https://helsinkisolutionoffice.atlassian.net/browse/TH-770)


Sartonon (approved):
Looking good

PR #113, opened 2020-09-11
TH-792 | Add aria-live element to tell the search status to SR users

jorilindell:
## Description
Add aria-live element to tell the search status to SR users

## Closes
[TH-792](https://helsinkisolutionoffice.atlassian.net/browse/TH-792)


Sartonon (approved):
Looks good!

PR #115, opened 2020-09-14
TH-782 | Accessibility/about page content from CMS

jorilindell:
## Description
Get about and accessibility pages from CMS
Remove static about pages

## Issues
[TH-782](https://helsinkisolutionoffice.atlassian.net/browse/TH-782)


src/domain/about/__tests__/AboutPage.test.tsx
```diff
@@ -1,27 +1,73 @@
...
   act(() => {
     i18n.changeLanguage('fi');
   });
 
-  expect(screen.getAllByText('Tietoa palvelusta')).toHaveLength(1);
+  expect(
+    screen.getAllByText(
+      fakeAboutPagesResponse.data[0].headingSection?.fi as string
+    )
+  ).toHaveLength(1);
```
Sartonon, line 37: is there a reason to not use screen.getByText()?

`expect(
    screen.getByText(
      fakeAboutPagesResponse.data[0].headingSection?.fi as string
    )
  ).toBeInTheDocument()`

Sartonon: if not, I would change this to make it simpler

src/domain/accessibility/AccessbilityPage.tsx
```diff
@@ -0,0 +1,56 @@
...
+      className={styles.accessibilityPage}
+      title="accessibility.title"
+    >
+      <MainContent offset={-70}>
+        <LoadingSpinner isLoading={loading}>
+          {headingSection && (
+            <Hero>
+              <div
+                dangerouslySetInnerHTML={{
+                  __html: sanitizeHtml(headingSection),
```
Sartonon, line 36: Hmm, is this sanitization needed here? The data come from trusted server? Or are there some other security concerns?

Maybe it is better to be safe than sorry :) but the downside is performance / sanitizeHtml package size

src/domain/about/__tests__/AboutPage.test.tsx
jorilindell: Simplified tests

src/domain/accessibility/AccessbilityPage.tsx
jorilindell: In this particular case the data is coming from our CMS. Same sanitize method is also used on Events page and I don't trust LinkedEvents data importers or actually source data so much. (So we need the package anyway) The data amount is not so huge so I don't see that performance is big problem here.

Sartonon: Ok sounds reasonable :)

Sartonon (approved):

PR #119, opened 2020-09-30
Helper function to get event fields

jorilindell:
## Description :sparkles:
Add helper function to get event fields
Use that helper function to unify way to get fields for components

### DIFF
--- a/src/domain/event/EventUtils.ts
+++ b/src/domain/event/EventUtils.ts
@@ -1,6 +1,7 @@
 import { isPast } from 'date-fns';
 import capitalize from 'lodash/capitalize';
 
+import { EVENT_STATUS } from '../../constants';
 import {
   EventFieldsFragment,
   PlaceFieldsFragment,
@@ -24,6 +25,15 @@ export const isEventClosed = (event: EventFieldsFragment): boolean => {
 };
 
 /**
+ * Check is event cancelled
+ * @param event
+ * @return {boolean}
+ */
+export const isEventCancelled = (event: EventFieldsFragment): boolean => {
+  return event.eventStatus === EVENT_STATUS.EVENT_CANCELLED;
+};
+
+/**
  * Check is event free
  * @param eventData
  * @return {boolean}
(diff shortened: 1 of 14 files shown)


src/domain/event/EventUtils.ts
```diff
@@ -164,20 +173,10 @@ export const getGoogleLink = (
...
-  );
-  const coordinates =
-    location && location.position
-      ? // Get coordinates for Google in correct order
-        [...location.position.coordinates].reverse()
-      : [];
+  const streetAddress = getLocalisedString(location?.streetAddress, locale);
+  const postalCode = location?.postalCode;
+  const addressLocality = getLocalisedString(location?.addressLocality, locale);
+  const coordinates = [...(location?.position?.coordinates || [])].reverse();
```
Sartonon, line 179: Maybe this line doesn't need the spread?

..?..?.reverse() || []

Sartonon (approved):
Good :)

src/domain/event/EventUtils.ts
jorilindell: I think that reverse function also mutate the original array, not only return value. I used spread syntax to avoid unpleasant side effects

Sartonon: yeah you are right :) didn't think of that, my bad!

then you could do:

location?.position?.coordinates?.slice().reverse()

but then  I think your solution is more readable

Sartonon: So I don't think this line needs any changes

jorilindell: This same line was used on 3 different functions. So I added a helper function to get event location fields

Sartonon: Yeah nice. Btw, why are coordinates always reversed?

jorilindell: LinkedEvents return coordinates in different order [lng, lat] than other services expect, so we have to change them to [lat, lng]

Sartonon: ok 👍

PR #121, opened 2020-10-01
Unify testing library

jorilindell:
## Description :sparkles:
Use @testing/react library whenever possible to write tests


src/domain/event/eventInfo/__tests__/OtherEventTimes.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+      start: event.startTime,
+      end: event.endTime,
+      locale: 'fi',
+      includeTime: true,
+      timeAbbreviation: translations.commons.timeAbbreviation,
+    });
+    expect(screen.getByText(dateStr)).toBeInTheDocument();
+  });
+
+  await waitFor(() => {
```
Sartonon, line 117: I think there is no point in having other stuff than

`expect(screen.getByText(dateStr)).toBeInTheDocument();`

inside this waitFor.Because that is the only thing that you are waiting for. Everything inside this will possibly be executed many times

src/domain/event/eventInfo/__tests__/OtherEventTimes.test.tsx
```diff
@@ -0,0 +1,177 @@
...
+      locale: 'fi',
+      includeTime: true,
+      timeAbbreviation: translations.commons.timeAbbreviation,
+    });
+    expect(screen.getByText(dateStr)).toBeInTheDocument();
+  });
+});
+
+test('should go to event page of other event time', async () => {
+  advanceTo(new Date('2020-08-11'));
```
Sartonon, line 142: Nice. I think this is better way to pass date to the function compared to what I have done

`new Date(2020, 7, 11)`

src/util/mockDataUtils.ts
```diff
@@ -32,58 +33,66 @@ export const fakeEvents = (
...
-    externalLinks: [] as any,
-    offers: [] as any,
-    subEvents: [] as any,
-    eventStatus: 'EventScheduled',
-    superEvent: null,
-    dataSource: 'hel',
-    __typename: 'EventDetails',
-    ...overrides,
-  };
+  return merge(
```
Sartonon, line 36: Why merge is needed?

Sartonon (changes requested):
Great stuff. I would at least fix the waitFor thing :)

src/util/mockDataUtils.ts
jorilindell: Using spread syntax for overrides caused me some extra work in cases when I wanted to replace only one field of nested object. 
For example if I want to override only name.fi not the complete name object.
```
const event = fakeEvent({
  name: { fi: eventName },
}) as EventFieldsFragment;
```

Sartonon: OK, nice addition 👍

src/domain/event/eventInfo/__tests__/OtherEventTimes.test.tsx
jorilindell: Very good point. I actually changed this so that I'm only waiting loading-spinner to disappear just to make it little bit more readable

Sartonon: Great!

Sartonon (approved):

PR #145, opened 2020-12-02
TH-883 | New routes/paths for selected places

Sartonon:
## Description :sparkles:

- Add new routes for selected places (annantalo etc.)

## Issues :bug:
**[TH-883](https://helsinkisolutionoffice.atlassian.net/browse/TH-883):** 

## Screenshots :camera_flash:
Screenshot 2020-12-02 at 18 14 03


src/domain/eventSearch/constants.ts
```diff
@@ -112,3 +112,13 @@ export const MAPPED_CATEGORIES: Record<string, string> = {
   [CATEGORIES.SPORT]: SPORT_KEYWORDS.join(','),
   [CATEGORIES.THEATRE]: 'yso:p2625',
 };
+
+export const MAPPED_PLACES: Record<string, string> = {
+  annantalo: 'tprek:7254',
```
nenonja, line 117: Where are these codes coming from `tprek:7256`. Could they change? Are these codes the same in every environment (test/production)?

Sartonon: https://api.hel.fi/linkedevents/v1/place/?division=kunta:helsinki&text=annantalo
https://api.hel.fi/linkedevents-test/v1/place/?division=kunta:helsinki&text=annantalo

They are id:s for the places. I figured them out just by searching the places from api with text. At leas Annantalo seems to be same id in both environments

Sartonon: Would be strange if those ids change

jorilindell: Those place ids are imported to Linked Events/Courses from tapahtumapaikkarekisteri and are same for all the environments. So this should be ok.

jorilindell (approved):
Looks good to me

PR #147, opened 2020-12-04
TH-861: start to use Helsinki Design System Navigation header component

nenonja:
## Description :sparkles:
This is the header part of the task. The footer part will come in next PR.

~~Work-in-progress:~~
- [x] Fix existing tests and add new ones for Header
- [x] Testcafe tests failing
- [x] Add screenshot here
- [x] Some colors may have changed

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-861

## Testing :alembic:
Verify that the header still works as expected (compare it to the production: https://tapahtumat.hel.fi/fi/home):
- [x] Navigation items are present and links work
- [x] In mobile view, navigation items can be opened from menubar (also with keyboard)
- [x] Languages are changeable (also with keyboard)
- [x] SkipToContent works with keyboard
- [ ] Check that colors are correct
- [ ] look and feel is same as in abstract: https://app.abstract.com/projects/b362e050-6671-11e9-9013-21682b9a514b/branches/master/commits/latest/files/72EAE992-755F-4ED8-8152-1F272C996F99/layers/CA33CFF9-2FBB-46CA-BBCC-755F5E383F84?collectionId=f7f4dedc-8bc9-4be7-b09c-99943a0d8aee&collectionLayerId=e36a6aa8-82cd-4b4a-9a3e-c025382630e5

## Screenshots :camera_flash:

Screenshot 2020-12-08 at 7 00 39 PM
Screenshot 2020-12-08 at 7 01 02 PM
Screenshot 2020-12-08 at 7 01 25 PM


README.md
```diff
@@ -24,11 +24,13 @@ Project is automatically deployed to testing environment when pushing to develop
 
 ### Install events-helsinki-api-proxy locally
 
-Clone the repository (https://github.com/City-of-Helsinki/events-helsinki-api-proxy). Follow the instructions for running open-city-profile with docker.
+Clone the repository (https://github.com/City-of-Helsinki/events-helsinki-ui). Follow the instructions for running open-city-profile with docker.
```
Sartonon, line 27: should be: https://github.com/City-of-Helsinki/events-helsinki-api-proxy

src/domain/app/header/header.module.scss
```diff
@@ -1,9 +1,33 @@
...
+      margin-right: var(--spacing-xs);
+    }
+  }
+  .navigationItemLabel {
+    display: flex;
+    align-items: center;
+  }
+
+  .languageSelector {
+    @media only screen and (max-width: $breakpoint-m) {
```
Sartonon, line 24: now would be good chance to add breakpoints.scss file to this project, and use mixins from there, also those breakpoints could be imported from design tokens, see:
https://github.com/City-of-Helsinki/linkedcomponents-ui/blob/develop/src/assets/styles/breakpoints.scss

Sartonon: or use the the mixins from layout.scss

Sartonon: here is a good example how those mixins are used:

https://github.com/City-of-Helsinki/events-helsinki-ui/blob/develop/src/domain/eventSearch/search.module.scss

Sartonon: so 2 options:

1. use breakpoints the same way linkedcomponents-ui uses
2. use the existing mixins from layout.scss (just add breakpoints from design tokens

README.md
nenonja: yes I read this lazily and thought is copy-paste error: "Clone this repository" etc...

package.json
```diff
@@ -177,6 +177,9 @@
       "src/util/testUtils.tsx",
       "src/util/updateSitemap.ts",
       "src/util/date-fns/locale/buildLocalizeFn.ts"
+    ],
+    "transformIgnorePatterns": [
+      "<rootDir>/node_modules/(?!hds-react)/"
```
Sartonon, line 182: Did only Navigation component have the problmen that required this? strange

nenonja: yes, only...

src/domain/app/header/header.module.scss
nenonja: layout.scss seem have a bit different breakpoint values than what comes from design tokens:
`$breakpoints: (
  xs: 576px,
  sm: 768px,
  md: 1024px,
  lg: 1200px,
  xlg: 1600px
);
`
Design tokens:
`$breakpoint-xs: 320px;
$breakpoint-s: 576px;
$breakpoint-m: 768px;
$breakpoint-l: 992px;
$breakpoint-xl: 1248px;`

Also layout.scss wants key value as parameter (xs, sm, md etc...) which does not go together with design tokens. 
So probably I go for same ways as linkedcomponents-ui does (1)

nenonja: introduced new breakpoints.scss, layout.scss can be deprecated

README.md
nenonja: fixed

package.json
jorilindell: I think the reason is that other HDS component are imported from `"hds-react"` not from `"hds-react/components...`

Sartonon: That would explain. Can you change the import?

Sartonon: @nenonja

src/domain/app/header/header.module.scss
Sartonon: This is very good now :)

Sartonon:
Screenshot 2020-12-09 at 15 20 08
Test failing for me, localization seems to be missing

src/domain/app/header/Header.tsx
```diff
@@ -1,15 +1,108 @@
-import React from 'react';
+import { Navigation } from 'hds-react/components/Navigation';
```
Sartonon: Change this import to look like this:

`import { Navigation } from 'hds-react';`

To avoid getting non compiled version

package.json
Sartonon: Tested this and changing the import style fixes the problem and this can be deleted

nenonja: Yes that did the trick: https://github.com/City-of-Helsinki/events-helsinki-ui/pull/147/commits/c8310040a6fb7dddc34015b6ddc2e67761a7fc32

testcafe/e2e/changeLanguage.ts
```diff
@@ -8,9 +10,16 @@ fixture('Landing page').page(getEnvUrl('/fi/home'));
 
 test('Changing language on landing page', async (t) => {
   await t
-    .expect(screen.getAllByRole('link', { name: /etsi tekemistä/i }).count)
+    .expect(
+      screen.getAllByRole('link', { name: translationsFi.header.searchEvents })
+        .count
+    )
     .eql(2)
```
Sartonon, line 17: expect(...).toHaveLength(); could be used here also. But this is minor, no need to change if you feel that .eql() is better

src/domain/app/header/Header.tsx
Sartonon: already fixed :)

src/domain/app/header/Header.tsx
```diff
@@ -1,15 +1,108 @@
...
   return (
-    <header className={styles.headerWrapper}>
-      <Navbar />
-      <MobileNavigation />
-    </header>
+    <Navigation
+      menuOpen={menuOpen}
+      onMenuToggle={onMenuToggle}
+      menuToggleAriaLabel={t('header.menuToggleAriaLabel')}
+      skipTo={`${location.pathname}${location.search}#${MAIN_CONTENT_ID}`}
```
jorilindell, line 68: It's not related to this PR. But it's little bit problematic that skipTo prop is just href of an anchor. I haven't tested this PR but I guess that after clicking "skip to content" link the page is totally reloaded and the Apollo cache is cleared. 

It would be nice to have onSkipTo where you could use react-router and prevent anchor event (same way as onTitleClick)

But as said not related to this PR.

Sartonon: I tested that now. Seems to work correctly, it doesn't refresh the page. Maybe it has something to do with the hash, because without it, it refreshes

Sartonon: https://stackoverflow.com/questions/4855168/what-is-href-and-why-is-it-used#:~:text=A%20hash%20%2D%20%23%20within%20a%20hyperlink,the%20id%20on%20that%20page.

explaned there. Hyperlink with hash only scrolls

Sartonon: BUT. I think i just figured that it only scrolls because the current url is same but you could also just pass the hash:

`skipTo={`#${MAIN_CONTENT_ID}`}`

jorilindell: Good, that it works. I guess you meant
```
skipTo={`#${MAIN_CONTENT_ID}`}
```

Sartonon: yeah I tried my best to escape those ticks but couldn't !!!! 🤯

testcafe/e2e/changeLanguage.ts
nenonja: Actually not, because it uses testcafe assertion. But I improved the test to check links both within footer and header separately.

src/domain/app/header/Header.tsx
nenonja: done

testcafe/e2e/changeLanguage.ts
Sartonon: Sorry, didn't event notice that this is actually testcafe test. my bad.

Sartonon (approved):

PR #149, opened 2020-12-06
Add malmitalo and fix back button functionaliy on event page

Sartonon:
## Description :sparkles:

- Add malmitalo and fix back button functionality on event page

## Issues :bug:
**[TH-902](https://helsinkisolutionoffice.atlassian.net/browse/TH-902):**

### DIFF
--- a/src/domain/event/eventCard/EventCard.tsx
+++ b/src/domain/event/eventCard/EventCard.tsx
@@ -31,6 +31,8 @@ const EventCard: React.FC<Props> = ({ event }) => {
   const history = useHistory();
   const { search } = useLocation();
   const { t } = useTranslation();
+  // place comes from place param in clean url on events search page
+  // see ROUTES.EVENT_PLACE (routes/constants.ts)
   const params = useParams<{ place?: string }>();
   const [showBackupImage, setShowBackupImage] = React.useState(false);
   const locale = useLocale();
--- a/src/domain/event/eventCard/EventCard.tsx
+++ b/src/domain/event/eventCard/EventCard.tsx
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import { IconArrowRight } from 'hds-react';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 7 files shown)


src/domain/event/eventCard/EventCard.tsx
```diff
@@ -30,6 +31,7 @@ const EventCard: React.FC<Props> = ({ event }) => {
   const history = useHistory();
   const { search } = useLocation();
   const { t } = useTranslation();
+  const params = useParams<{ place?: string }>();
```
jorilindell, line 34: Could be useful to mention here why place is got from the url parameter. It's not 100% clear that place is used only on the place specific paces. Can be little bit confusing to new developer in the team.

src/domain/event/eventCard/LargeEventCard.tsx
```diff
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import { Button } from 'hds-react';
```
jorilindell, line 2: This is not related to this PR. But there are quite much same logic on  EventCard and LargeEventCard. The layout and buy button are pretty much only differences. I know this was my decision but it could be good idea to have all the logic in single component. Could be little bit easier to maintain.

src/domain/event/eventCard/LargeEventCard.tsx
```diff
@@ -30,6 +31,7 @@ interface Props {
 const LargeEventCard: React.FC<Props> = ({ event }) => {
   const { t } = useTranslation();
   const { push } = useHistory();
+  const params = useParams<{ place?: string }>();
```
jorilindell, line 34: Could be useful to mention here why place is got from the url parameter. It's not 100% clear that place is used only on the place specific paces. Can be little bit confusing to new developer in the team.

jorilindell (approved):
Looks good. Added some quite minor improvement ideas.

src/domain/event/eventCard/EventCard.tsx
Sartonon: true

PR #150, opened 2020-12-07
TH-903, TH-905 | Add missing path from sitemap generator + add meta data

Sartonon:
## Description :sparkles:

- Add metadata to collections page

- Add metadata to events search page

- Add missing paths to sitemaps generator

## Issues :bug:
**[TH-903](https://helsinkisolutionoffice.atlassian.net/browse/TH-903):** 
**[TH-905](https://helsinkisolutionoffice.atlassian.net/browse/TH-905):**

### DIFF
--- a/src/util/updateSitemap.ts
+++ b/src/util/updateSitemap.ts
@@ -3,6 +3,8 @@ import range from 'lodash/range';
 import { promisify } from 'util';
 import * as convert from 'xml-js';
 
+import { MAPPED_PLACES } from '../domain/eventSearch/constants';
+
 export type Language = 'en' | 'fi' | 'sv';
 export type TitleKey = 'title_en' | 'title_fi' | 'title_sv';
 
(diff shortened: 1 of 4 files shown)


src/util/updateSitemap.ts
```diff
@@ -36,7 +38,9 @@ export type Element = {
 const writeFile = promisify(fs.writeFile);
 
 const LANGUAGES = ['en', 'fi', 'sv'];
-const STATIC_URLS = ['home'];
+const STATIC_URLS = ['home', 'events', 'collections'].concat(
+  Object.keys(MAPPED_PLACES)
+);
```
jorilindell, line 43: To me `const STATIC_URLS = ['home', 'events', 'collections', ...Object.keys(MAPPED_PLACES)];` would be easier to read. But it's really just matter of taste

jorilindell (approved):

src/util/updateSitemap.ts
Sartonon: .concat came first to my mind because I saw it used somewhere recently, I agree that spread is better here

PR #152, opened 2020-12-10
Fix HDS css color variable names

jorilindell:


nenonja (approved):
Looks good but I did not test this yet

Sartonon (approved):
Tested and seems to work! :)

PR #157, opened 2020-12-15
Fix division in event search

nenonja:
## Description :sparkles:

Fix problem that division is ignored from event search. Also use always localOngoingSearch when division is selected in order to remove onlineEvents from division searches.
Tried to write regression test but it went too complicated because of complex apollo mocks that wont fit on this scenario.

### DIFF
--- a/src/domain/eventSearch/utils.tsx
+++ b/src/domain/eventSearch/utils.tsx
@@ -16,7 +16,7 @@ import isNumber from 'lodash/isNumber';
 import React from 'react';
 
 import { DATE_TYPES } from '../../constants';
-import { Meta } from '../../generated/graphql';
+import { Meta, QueryEventListArgs } from '../../generated/graphql';
 import IconCultureAndArts from '../../icons/IconCultureAndArts';
 import IconDance from '../../icons/IconDance';
 import IconFood from '../../icons/IconFood';
@@ -175,7 +175,7 @@ export const getEventSearchVariables = ({
   sortOrder: EVENT_SORT_OPTIONS;
   superEventType: string[];
   place?: string;
-}) => {
+}): QueryEventListArgs => {
   const {
     categories,
     dateTypes,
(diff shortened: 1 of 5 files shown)


src/domain/eventSearch/utils.tsx
```diff
@@ -221,11 +221,13 @@ export const getEventSearchVariables = ({
...
     .filter((e) => e);
 
+  const hasDivisions = !isEmpty(divisions);
+  const hasText = !isEmpty(text);
   // Combine and add keywords
-
   return {
-    allOngoingAnd: text,
-    ...(!isEmpty(divisions) && { division: divisions.sort() }),
+    ...(hasText &&
```
nenonja, line 228: complex code but it ensures that the parameters are added only when they have nonempty value. This makes also mock writing easier when we dont have to expect null or undefined value or empty list.

Sartonon: if his gets hard to read we could just create object in a variable and add these fields in if statements and then return the object

Sartonon (approved):
Looks good

src/domain/eventSearch/utils.tsx
jorilindell: Should internet events also be hidden when there is at least one place selected?

jorilindell: And maybe also `division: ["kunta:helsinki"]` should be added to query when no other division is selected and `text` array is empty.

nenonja: > Should internet events also be hidden when there is at least one place selected?

Yes good idea I added the check

nenonja: > And maybe also `division: ["kunta:helsinki"]` should be added to query when no other division is selected and `text` array is empty.

I think it's redundant data because when `allOngoingAnd` is added, it searches from helsinki area by default and seemed to completely ignore division values.

Sartonon (approved):

PR #160, opened 2020-12-16
TH-892 | Single course page

Sartonon:
## Description :sparkles:

- Added single course page

## Issues :bug:
**[TH-892](https://helsinkisolutionoffice.atlassian.net/browse/TH-892):** 

## Screenshots :camera_flash:


Sartonon (commented):
This one has been long time in the works :D Most likely has much to improve, need to review it myself too now :D

src/domain/app/routes/constants.ts
```diff
@@ -1,16 +1,15 @@
...
 // Remember also update static urls to updateSitemap
-
 export type ROUTE_TYPES = {
   ABOUT: '/about';
   ACCESSIBILITY: '/accessibility';
   COLLECTION: '/collection/:slug';
   COLLECTIONS: '/collections';
   EVENTS: '/events';
-  EVENT: '/event/:id';
+  EVENT: '/events/:id';
```
jorilindell, line 10: This change can cause some problems if there are already some shared events e.g. in facebook. After this change they will open 404 page. You could add redirect from '/event/:id' to '/events/:id' to avoid this problem. Of course the problem is only temporary

src/domain/event/eventInfo/EventInfo.tsx
```diff
@@ -5,19 +5,20 @@ import {
   IconAngleRight,
```
jorilindell, line 5: In this file there is 
```
const downloadIcsFile = () => {
    if (startTime) {
      const domain = getDomain();
      const icsEvent: EventAttributes = {
        description: t('event.info.textCalendarLinkDescription', {
          description: shortDescription,
          link: `${domain}/${locale}${ROUTES.EVENT.replace(':id', event.id)}`,
        }),
        end: endTime ? getDateArray(endTime) : getDateArray(startTime),
        location: [locationName, streetAddress, district, addressLocality]
          .filter((e) => e)
          .join(', '),
        productId: domain,
        start: getDateArray(startTime),
        startOutputType: 'local',
        title: name,
      };
      createEvent(icsEvent, (error: Error | undefined, value: string) => {
        if (error) {
          Sentry.captureException(error);
        } else {
          const blob = new Blob([value], { type: 'text/calendar' });
          saveAs(blob, `event_${event.id.replace(/:/g, '')}.ics`);
        }
      });
    }
  };
```
function which now uses event url also for courses in description link

.env
```diff
@@ -6,6 +6,7 @@ REACT_APP_GRAPHQL_BASE_URL=https://tapahtumat-proxy.test.kuva.hel.ninja/proxy/gr
 
 REACT_APP_CMS_URL=https://cms.test.kuva.hel.ninja/api
 REACT_APP_LINKED_EVENTS_URL=https://api.hel.fi/linkedevents/v1
+REACT_APP_LINKED_COURSES_URL=https://api.hel.fi/linkedcourses/v1
```
jorilindell, line 9: In think this variable should also be added to Dockerfile

src/util/updateSitemap.ts
```diff
@@ -268,10 +288,10 @@ const getEventsPage = async (start: Date, page: number) => {
  * @param {string} start
  * @return {Object[]}
  */
-const getEvents = async (start: Date) => {
+const getEvents = async (eventType: 'event' | 'course', start: Date) => {
```
jorilindell, line 291: You could also fetch all internet based events/courses. Example request: https://api.hel.fi/linkedevents/v1/event?start=now&internet_based=True&&super_event_type=umbrella,none. Now they are not added to sitemap

jorilindell:
This is design issue, but course page links are not passing contrast check, which can cause some problems in accessibility audit
Screenshot 2020-12-23 at 12 35 43

jorilindell:
After clicking back button in course page the course list is not scrolled to the right position. This can be reproduced like this:
1. Open course search page e.g. http://localhost:3000/fi/courses/harrastushaku:11401_2020122314001500?text=taide
2. Scroll little bit down and click a course card (not the first one)
3. In course page click back button
Error: Browser is not automatically scrolled to the right course card. In event search page this is working and was implemented by storing event id to react-router state when clicking back button. In event search page there is function that check the react-router state and scrolls to right position and event id is stored

nenonja (commented):
Something wrong at the front page
Screenshot 2020-12-23 at 3 08 30 PM

nenonja:
Front page hides the search under header
Screenshot 2020-12-23 at 3 12 13 PM

nenonja:
For some reason I cannot find any search results when selecting some division:
Screenshot 2020-12-23 at 3 21 26 PM
This is the search on api-proxy:
 https://api.hel.fi/linkedcourses/v1/event?division=kaupunginosa:lauttasaari&include=keywords,location&language=fi&page_size=10&sort=end_time&start=now&super_event_type=umbrella,none (602ms)

nenonja (commented):
Lots of espoo area results are on the top when opening course search page, or search with:
 https://api.hel.fi/linkedcourses/v1/event?include=keywords,location&language=fi&page_size=10&sort=end_time&start=now&super_event_type=umbrella,none

jorilindell:
> Front page hides the search under header
> Screenshot 2020-12-23 at 3 12 13 PM

To me it looks that GraphQl proxy is not getting landing page data from CMS and hero content is missing

src/domain/event/EventUtils.ts
```diff
@@ -189,6 +190,15 @@ const getEventLocationFields = (
   };
 };
 
+const getCourseFields = (event: EventFields) => {
```
nenonja, line 193: Why is this under EventUtils.ts?

src/domain/app/routes/constants.ts
Sartonon: good to know, maybe I will add some kind of redirection for now

src/domain/event/EventUtils.ts
Sartonon: hmm, maybe because it is used here in the getEventsFields. Not sure what is the most logical place for this

Sartonon:
> After clicking back button in course page the course list is not scrolled to the right position. This can be reproduced like this:
> 
> 1. Open course search page e.g. http://localhost:3000/fi/courses/harrastushaku:11401_2020122314001500?text=taide
> 2. Scroll little bit down and click a course card (not the first one)
> 3. In course page click back button
>    Error: Browser is not automatically scrolled to the right course card. In event search page this is working and was implemented by storing event id to react-router state when clicking back button. In event search page there is function that check the react-router state and scrolls to right position and event id is stored

yeah actually scrolling part still has TODO comment. Maybe I will add scrolling to this PR

Sartonon:
> course search page has misplaced dropdown
> Screenshot 2020-12-23 at 3 08 30 PM

I think it is in the correct place: see https://app.abstract.com/projects/b362e050-6671-11e9-9013-21682b9a514b/branches/master/commits/latest/files/72EAE992-755F-4ED8-8152-1F272C996F99/layers/CA33CFF9-2FBB-46CA-BBCC-755F5E383F84?collectionId=f7f4dedc-8bc9-4be7-b09c-99943a0d8aee&collectionLayerId=e36a6aa8-82cd-4b4a-9a3e-c025382630e5

src/domain/course/CoursePageContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { useSimilarCoursesQuery } from '../event/queryUtils';
+import SimilarEvents from '../event/similarEvents/SimilarEvents';
+import { EventFields, EventType } from '../event/types';
+import styles from './coursePage.module.scss';
+
+interface RouteParams {
+  id: string;
+}
+
+const CoursePageContainer: React.FC = () => {
```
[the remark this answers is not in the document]
nenonja: or just broaden the usage of EventPageContainer for courses too. It seems that `event` is common name for any event, and course is just special type of event.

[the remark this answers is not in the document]
Sartonon: true, I will try to come up with component that has at least the common jsx, so the that is not duplicated

src/util/updateSitemap.ts
Sartonon: good catch!

.env
Sartonon: 👍

src/domain/event/queryUtils.ts
```diff
@@ -0,0 +1,224 @@
...
+      pageSize: PAGE_SIZE,
+      params: searchParams,
+      sortOrder: EVENT_SORT_OPTIONS.END_TIME,
+      superEventType: ['umbrella', 'none'],
+    });
+  }, [locale, searchParams]);
+};
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const useSimilarEventsQuery = (event: EventFields) => {
```
nenonja, line 50: Could this be combined with `useSimilarCoursesQuery` ? 
```suggestion
export const useSimilarEventsQuery = (event: EventFields, useQuery = useEventListQuery, listType = 'eventList') => { ... }
```

src/domain/event/queryUtils.ts
```diff
@@ -0,0 +1,224 @@
...
+      superEvent: superEventId,
+    }),
+    [superEventId]
+  );
+
+  return { superEventId, variables };
+};
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const useOtherEventTimes = (event: EventFields) => {
```
nenonja, line 109: Could this be combined with `useOtherCourseTimes` ? 
```suggestion
export const useOtherEventTimes = (event: EventFields, useQuery = useEventListQuery, listType = 'eventList') => { ... }
```

nenonja (commented):
Most radical refactoring idea would be to use only existing events components and functionality and broaden it to use courses. Later these will be combined on backend side, so we could start to do the same instantly? Also if we decide to introduce feature flag for courses later, it is easier when functionality is unified.
One last note, I think that this pr is so super big that it could be merged to events-helsinki-ui instantly and do possible review fixes in a new smaller pr. That reduces chance for new merge conflict pains.

nenonja:
> > course search page has misplaced dropdown
> > Screenshot 2020-12-23 at 3 08 30 PM
> 
> I think it is in the correct place: see https://app.abstract.com/projects/b362e050-6671-11e9-9013-21682b9a514b/branches/master/commits/latest/files/72EAE992-755F-4ED8-8152-1F272C996F99/layers/CA33CFF9-2FBB-46CA-BBCC-755F5E383F84?collectionId=f7f4dedc-8bc9-4be7-b09c-99943a0d8aee&collectionLayerId=e36a6aa8-82cd-4b4a-9a3e-c025382630e5

I think that the idea in abstract is to place first dropdown to the end of first row when there is odd amount of dropdowns. Now there is even amount of dropdowns and thus looks funny. But I think it's going to look ok in the end if we have then all the dropdowns implemented

src/util/updateSitemap.ts
jorilindell: You can also wait until `all_ongoing=true` parameter is implemented and get all events using just one request. https://helsinkicity.slack.com/archives/C019AV8SKA9/p1608734354007800

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </button>
+      </InfoWithIcon>
+      {isListOpen && (
+        <>
+          <ul
+            className={styles.timeList}
+            data-testid={otherEventTimesListTestId}
+          >
+            {events.map((event) => {
+              const moveToEventPage = () => {
```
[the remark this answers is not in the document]
Sartonon: Moved functions outside, but can't see why react.memo would be useful here, and? if you meant useCallback, I think it only adds overhead here, no perf benefits

src/domain/event/queryUtils.ts
Sartonon: I think I initially thought something like this, but my biggest concern was that handling the types would be hard.

Sartonon: I can test if I can come up with some solution with the typings. Maybe I will learn something new :D

src/domain/event/queryUtils.ts
Sartonon: With this I thought that the typing would be even harder, but maybe I need to try something to find out if it is doable with good typings

Sartonon:
```
export const useOtherEventTimes = (
  event: EventFields,
  useQuery: typeof useEventListQuery | typeof useCourseListQuery
) => {
```

Maybe something like that and some kind of type guarding will make this possible

Sartonon: Or if you have other ideas to handle typing, please tell me :)

package.json
```diff
@@ -73,7 +73,7 @@
     "react-toastify": "^6.0.9",
     "sanitize-html": "^1.20.1",
     "serialize-javascript": "^3.1.0",
-    "typescript": "3.9.3",
+    "typescript": "4.1.2",
```
nenonja, line 76: Got this: Maybe typescript-eslint should be updated also? Preferably I would add version bumps in separate pr.
`WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.
You may find that it works just fine, or you may not.
SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <4.1.0
YOUR TYPESCRIPT VERSION: 4.1.2
`

src/util/updateSitemap.ts
Sartonon: all_ongoing=true seems to return less events

Sartonon:
> This is design issue, but course page links are not passing contrast check, which can cause some problems in accessibility audit
> Screenshot 2020-12-23 at 12 35 43

hmm that is already darkest fog color available, need to ask Tomi what color we should use

src/domain/app/routes/constants.ts
Sartonon: fixed

.env
Sartonon: added

src/domain/course/CoursePageContainer.tsx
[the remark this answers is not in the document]
Sartonon: refactored

package.json
Sartonon: downgraded back for now

Sartonon:
> This is design issue, but course page links are not passing contrast check, which can cause some problems in accessibility audit
> Screenshot 2020-12-23 at 12 35 43

Link colors changed for course page to have better conrast

src/domain/course/CoursePageContainer.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+  return (
+    <PageWrapper className={styles.coursePageWrapper} title="event.title">
+      <MainContent offset={-70}>
+        <LoadingSpinner isLoading={loading}>
+          {course ? (
+            <>
+              {/* Wait for data to be accessible before updating metadata */}
+              <EventPageMeta event={course} />
+              {courseClosed ? (
```
nenonja, line 54: This could be simpler?
```suggestion
             {courseClosed ? (
                <EventClosedHero />
              ) : (
                <EventHero event={course} eventType={EventType.COURSE} />
                <EventContent event={course} eventType={EventType.COURSE} />
              )}
```

Sartonon: hmm CoursePageContainer and EventPageContainer should be maybe combined to use same codem like I just did for the searchPage

because this jsx part is copy paste from EventPageContainer

Sartonon: But you are right, your suggestion is better, good catch

src/domain/event/eventCard/__tests__/EventCard.test.tsx
```diff
@@ -57,15 +57,16 @@ test('should go to event page by clicking event card', () => {
   expect(history.location.pathname).toEqual('/');
 
   userEvent.click(
-    screen.queryByRole('link', {
+    screen.getByRole('link', {
       name: translations.event.eventCard.ariaLabelLink.replace(
         '{{name}}',
-        event.name.fi
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
nenonja, line 63: is this necessary? When I remove this `as any`change it still worked

src/domain/event/eventCard/__tests__/EventCard.test.tsx
```diff
@@ -57,15 +57,16 @@ test('should go to event page by clicking event card', () => {
...
+        event.name.fi as any
       ),
     })
   );
 
-  expect(history.location.pathname).toEqual('/fi/event/123');
```
nenonja, line 68: as the backward compability was left, we could check here both `event/` and `events/` ?

src/domain/event/eventHero/__tests__/EventHero.test.tsx
```diff
@@ -47,8 +47,12 @@ afterAll(() => {
   clear();
 });
 
+const renderComponent = (props?: Partial<EventHeroProps>) => {
```
nenonja, line 50: props seems to be redundant (it only could have `event` and `eventType` which are already set)

nenonja: also you could use `eventType={EventType.EVENT}`

src/domain/event/eventCard/__tests__/EventCard.test.tsx
Sartonon: Screenshot 2020-12-31 at 11 29 00
Screenshot 2020-12-31 at 11 28 52

I think we should override that rule for test files anyway

src/domain/event/eventCard/__tests__/EventCard.test.tsx
Sartonon: yeah maybe, but that is a totally different test in a different place?

src/domain/event/eventHero/__tests__/EventHero.test.tsx
Sartonon: > props seems to be redundant (it only could have `event` and `eventType` which are already set)

I don't they are redundant, the idea is that you can override the default props that are set in the function for the component



> also you could use `eventType={EventType.EVENT}`

true, changed/fixed

src/domain/event/eventInfo/EventInfo.tsx
```diff
@@ -123,7 +156,11 @@ const EventInfo: React.FC<Props> = ({ event }) => {
         </InfoWithIcon>
 
         {/* Other event times */}
-        <OtherEventTimes event={event} />
+        {eventType === EventType.EVENT ? (
+          <OtherEventTimesContainer event={event} />
```
nenonja, line 160: eventType could be passed as parameter, so we should not need separate `OtherEventTimesContainer` and `OtherCourseTimesContainer` components?

src/domain/event/eventHero/__tests__/EventHero.test.tsx
Sartonon: Screenshot 2020-12-31 at 11 35 57

no params are passed in any test atm but if we would add more tests and override the props the typing is useful

src/domain/event/eventInfo/EventInfo.tsx
```diff
@@ -147,6 +184,18 @@ const EventInfo: React.FC<Props> = ({ event }) => {
           </Link>
         </InfoWithIcon>
 
+        {/* Audience */}
```
nenonja, line 187: There is new components showing in EventInfo.tsx (otherCourses, audience items, courseInfo rows) but these are not covered in EventInfo.test.tsx?
When the coverage is covered to pr, we could easily see that coverage has dropped for this component.

src/domain/event/EventUtils.ts
```diff
@@ -281,17 +291,35 @@ const getOfferInfoUrl = (
   return getLocalisedString(offer?.infoUrl, locale);
 };
 
+export const getEventAudience = (
```
nenonja, line 294: this is quite complex method but not covered with unit tests?

src/domain/event/EventUtils.ts
```diff
@@ -281,17 +291,35 @@ const getOfferInfoUrl = (
...
+      id: audience.id || '',
+      name: capitalize(audience.name?.[locale] || '').trim(),
+    }))
+    .filter(
+      (audience, index, arr) =>
+        !!audience.id &&
+        !!audience.name &&
+        !EVENT_KEYWORD_BLACK_LIST.includes(audience.id) &&
+        arr.findIndex(
+          (item) => item.name.toLowerCase() === audience.name.toLowerCase()
```
nenonja, line 309: no need to make lower case if all item names are capitalized? (line 301)

src/domain/event/eventCard/__tests__/LargeEventCard.test.tsx
```diff
@@ -102,7 +102,7 @@ test('should go to event page by click Read more button', () => {
     })
   );
 
-  expect(history.location.pathname).toEqual('/fi/event/123');
```
nenonja, line 105: if backward compability was left, we could check both `event` and `events`?

src/domain/event/eventInfo/EventInfo.tsx
Sartonon: No sure how that would work

Sartonon: These containerr components are where because of the different hooks

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
```diff
@@ -0,0 +1,106 @@
+import { IconAngleDown, IconArrowRight, IconCalendarPlus } from 'hds-react';
```
nenonja, line 1: Files seems to be just moved but git shows it as new file which makes reviewing harder. `git mv` or just `mv` file then git will show it as moved file.

src/domain/event/types.ts
```diff
@@ -1,4 +1,27 @@
...
+  EVENT = 'event',
+  COURSE = 'course',
+}
+
+export const EVENT_ROUTE_MAPPER = {
+  [EventType.EVENT]: ROUTES.EVENT,
+  [EventType.COURSE]: ROUTES.COURSE,
+};
+
+export const EVENTS_ROUTE_MAPPER = {
```
nenonja, line 24:
```suggestion
export const EVENTS_ROUTE_MAPPER = EVENT_ROUTE_MAPPER;
```

src/domain/event/eventCard/__tests__/LargeEventCard.test.tsx
Sartonon: yeah maybe, but that is a totally different test in a different place?

Sartonon: that compability has nothing to do with LargeEventCard functionality

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
Sartonon: oh, never used those. Good to know in future

src/domain/event/types.ts
Sartonon: I don't understand this suggestion.

Sartonon: They are not quite the same

src/domain/event/types.ts
```diff
@@ -1,4 +1,27 @@
...
+import { ROUTES } from '../app/routes/constants';
+
 export type KeywordOption = {
   id: string;
   name: string;
 };
+
+export type EventFields = EventFieldsFragment | CourseFieldsFragment;
+
+export enum EventType {
```
nenonja, line 14: Using just type might be easier than enum?
```suggestion
export type EventType = 'event' | 'course';

export const EVENT_ROUTE_MAPPER: Record<EventType, string> = {
  event: ROUTES.EVENT,
  course: ROUTES.COURSE,
};

```

src/domain/courseSearch/utils.ts
```diff
@@ -0,0 +1,78 @@
+import { forEach, isArray, isEmpty, isNil, isNumber } from 'lodash';
+
+import { Meta } from '../../generated/graphql';
+import { formatDate } from '../../util/dateUtils';
+import getUrlParamAsArray from '../../util/getUrlParamAsArray';
+import { EVENT_SEARCH_FILTERS } from '../eventSearch/constants';
+import { Filters, MappedFilters } from './types';
+
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const getSearchFilters = (searchParams: URLSearchParams) => {
```
nenonja, line 10: My IDE says that none of these methods are never used anywhere in the codebase. Accidental or is this code that will be used in the future? No test coverage either.

src/domain/event/EventUtils.ts
Sartonon: well, true, but do we want to delete them really? I think these should be left here for safety reasons if for some reason capitalization was moved somewhere else or deleted? If capitalization was deleted, you wouldn't notice that toLowerCase needs to be added

nenonja (commented):
Some comments added again... one option would be to merge this to develop and continue fixes in another pr if needed?

src/domain/event/EventUtils.ts
Sartonon: but of course good tests would catch that error

Sartonon: Trying to figure out where this function event came from to this pull request.. I don't remember writing this code at all

src/domain/course/CoursePageContainer.tsx
nenonja: > hmm CoursePageContainer and EventPageContainer should be maybe combined to use same codem like I just did for the searchPage

Yes I was about to suggest that but thinked that you mentioned previously that some of the refactoring will be left later or so... Anyway combining these two would be great.

src/domain/event/eventInfo/EventInfo.tsx
nenonja: hook is just a function, you can use different function depending on the event param?
`const {field1, field2 } = param ? useHook1(event) : useHook2(event); `

src/domain/event/EventUtils.ts
nenonja: I think it's better to add unit tests (for upper, lower etc) and then simplify code. Unittest it the best safety guard imho.

src/domain/event/eventCard/__tests__/EventCard.test.tsx
nenonja: I read this very badly, it's about the history here. I thought that we just followed the route. But there could be a test in AppRoute or somewhere. No need to do that if it has too much work.

src/domain/event/types.ts
nenonja: I was just blind! Too many hours reviewing 🤓

src/domain/courseSearch/utils.ts
Sartonon: Hmm, I think that this might have already been deleted in some commit but came back in some rebase or merge... not sure

Anyway, now deleted

src/domain/event/types.ts
Sartonon: it is expected to be enum, it is also used in other places:

Screenshot 2020-12-31 at 15 01 14

src/domain/event/EventUtils.ts
Sartonon: Well, I agree that it is better to add unit tests anyway. But I wouldn't rely 100% the tests to be correct. .toLowerCase() is not that complicated, just a little verbose

src/domain/event/eventInfo/EventInfo.tsx
Sartonon: more testing coverage added, other course or other event are not tested here as integration tests, but those components have tests  in itself

src/domain/event/types.ts
nenonja: Just put `eventType="course"` there? If you put illegal string value there, ts will complain.

src/domain/event/EventUtils.ts
Sartonon: combined with the other function and improved tests to have the casing tested

src/domain/event/EventUtils.ts
Sartonon: refactored + improved tests and decided to get rid of toLowerCase

src/domain/event/types.ts
Sartonon: hmm good point, that might be better solution

src/domain/event/eventInfo/EventInfo.tsx
Sartonon: hooks cannot be used conditionally

Sartonon: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

nenonja: Ok forgot about that but one option is to put the condition inside custom hook as I think that hook can call another hook conditionally without react complaining

Sartonon: And how do you make sure that it is not gonna break now or in the future?

Sartonon: and putting the condition inside another custom hooks is not gonna change anything, react still complains

nenonja: Could there be a chance to use skip-value? https://stackoverflow.com/questions/59489097/use-query-hook-react-apollo-conditional-call
Not sure how to do this, genreally there should be a way to avoid copy-pasting (now components are 95% same)
Maybe it can be done much easier when courses and events are unified on backend. And this can be resolved & merged

Sartonon: They could be moved to same file to avoid some repetition with imports but basically the repetition is one line (using component with different props).

maybe skip value could be used but I think it would make the component overly complicated (which props are passed from hooks to the component?), and if more EventTypes would be added itwould be event more complicated.

And of course I would like to wait for the new linked-events API. Maybe then we can come up with better implementation with fetching events and courses.

And we are basically creating helper-components that use different hooks and pass different props to the other component, I would harly call that "copy-paste".

Something to read:
https://kentcdodds.com/blog/aha-programming
https://medium.com/@thisdotmedia/the-cost-of-premature-abstraction-b5d71ffd6400

PR #161, opened 2020-12-16
Add husky and lint-staged configuration

melniiv:
-Uses linter and prettier before committing changes.
-Runs tests automatically before pushing the changes.
-Fixes the existing lint errors and warnings.

### DIFF
--- a/.huskyrc
+++ b/.huskyrc
@@ -1,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "lint-staged"
+  }
+}
--- a/.lintstagedrc
+++ b/.lintstagedrc
@@ -1,0 +1,7 @@
+{
+  "*.{js,jsx,tsx}": [
+    "yarn format-code",
+    "yarn test:staged",
+    "prettier --write",
+  ]
+}
--- a/package.json
+++ b/package.json
@@ -86,13 +86,14 @@
     "build:server": "NODE_ENV=production webpack --progress --mode production --config config/webpack.server.config.js",
     "codegen": "graphql-codegen -r dotenv/config dotenv_config_path=.env.development.local  --config codegen.yml",
     "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
+    "test:staged": "react-scripts test --env=jest-environment-jsdom-sixteen --watchAll=false --findRelatedTests",
     "test:coverage": "react-scripts test --coverage --env=jest-environment-jsdom-sixteen --watchAll=false",
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
     "browser-test": "testcafe \"chrome --window-size='1920,1080'\" testcafe/e2e/",
     "browser-test:local": "BROWSER_TEST_ENV=local yarn browser-test",
     "browser-test:ci": "testcafe \"chrome:headless --disable-gpu --window-size='1920,1080'\" testcafe/e2e/",
     "lint": "eslint --ext js,ts,tsx src",
-    "format-code": "eslint --ext js,ts,tsx src --fix"
+    "format-code": "eslint --ext js,ts,tsx src --fix --max-warnings=0"
   },
   "eslintConfig": {
     "extends": "react-app"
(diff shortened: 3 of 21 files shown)


package.json
```diff
@@ -86,6 +86,7 @@
     "build:server": "NODE_ENV=production webpack --progress --mode production --config config/webpack.server.config.js",
     "codegen": "graphql-codegen -r dotenv/config dotenv_config_path=.env.development.local  --config codegen.yml",
     "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
+    "test:all": "set CI=true && react-scripts test --env=jest-environment-jsdom-sixteen --watchAll=false && yarn lint",
```
Sartonon, line 89: maybe we should lint first, so we can fail faster? Tests take pretty long

src/common/components/dateRangePicker/__tests/DateRangePicker.test.tsx
```diff
@@ -148,7 +148,7 @@ test('should call onChangeStartDate with clicking date', async () => {
   });
   userEvent.click(startDateInput);
 
-  userEvent.click(screen.getByRole('option', { name: /day\-15/i }));
+  userEvent.click(screen.getByRole('option', { name: /day-15/i }));
```
Sartonon, line 151: do these tests still work? was backslash redundant?

src/domain/eventSearch/__tests__/EventSearchPageContainer.test.tsx
```diff
@@ -263,4 +264,4 @@ it('should scroll to result list on mobile screen', async () => {
   expect(scroller.scrollTo).toBeCalled();
 });
 
-it('searches events correctly with event place in path, e.g. /fi/annantalo', () => {});
+//it('searches events correctly with event place in path, e.g. /fi/annantalo', () => {});
```
Sartonon, line 267: hmm seems like I forgot this here earlier :D

.huskyrc
```diff
@@ -0,0 +1,6 @@
+{
+  "hooks": {
+    "pre-commit": "lint-staged",
```
Sartonon, line 3: in pre-commit we could test the related files with jest

.huskyrc
```diff
@@ -0,0 +1,6 @@
...
+    "pre-push": "yarn test:all"
```
Sartonon, line 4: Maybe we should leave this to the pipeline? It takes painfully long to run test for all files in every push :/

src/common/components/dateRangePicker/__tests/DateRangePicker.test.tsx
melniiv: yes they pass. i think its safe.

src/domain/eventSearch/__tests__/EventSearchPageContainer.test.tsx
melniiv: i didnt remove, so if needed, it can be implemented later, but still present in the code

.huskyrc
melniiv: :) then the whole PR is useless. Yes, it takes time, maybe couple of minutes more.. but then its just double check, that repo always has the "right" and working version. Pipeline can double check of course.. because its not a 100% way to restrict pushing the bad code.

.huskyrc
melniiv: i dont think i got what you mean. example?

Sartonon: by default jest tests only files that have changed

Sartonon: and all related files

.huskyrc
Sartonon: I mean we could run the tests pre-commit for staged files

src/domain/eventSearch/__tests__/EventSearchPageContainer.test.tsx
Sartonon: 👍

.huskyrc
Sartonon: or maybe it was default for react-scripts test:

see here: https://github.com/City-of-Helsinki/palvelutarjotin-ui/blob/develop/package.json

Sartonon: ` "test:staged": "SASS_PATH=./src/assets/styles jest --findRelatedTests",`

package.json
melniiv: yea, we can

.lintstagedrc
```diff
@@ -0,0 +1,7 @@
+{
+  "*.{js,jsx,tsx}": [
+    "yarn lint",
+    "yarn test:staged",
+    "prettier --write",
+  ]
+}
```
Sartonon, line 7: Good! Just realised that lint --fix is used in palvelutarjotin, Should we use that also here? For example import ordering can be easily fixed with the current configuration here

Sartonon (approved):
Looks great 👍

PR #163, opened 2020-12-19
TH-938 | Update Dockerfile to generate smaller image

Sartonon:
## Description :sparkles:

- Update Dockerfile to generate smaller image

Production build seems to have one external dependency (canvas). We can reduce the final image size by just installing that package in the image. Seems to work.

There seemed to be some errors if you try to remove the external dependecy from webpack config.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ RUN yarn policies set-version $YARN_VERSION
 USER appuser
 
 # Install dependencies
-COPY --chown=appuser:appuser package*.json *yarn* /app/
+COPY --chown=appuser:appuser package.json *yarn* /app/
 RUN yarn && yarn cache clean --force
 
 # Copy all files
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,9 +34,10 @@ ENV NODE_ENV $NODE_ENV
 # Bake package.json start command into the image
 CMD ["yarn", "start"]
 
-# ==========================================
-FROM appbase as production
-# ==========================================
+# =============================
+FROM appbase as prodbuilder
+# =============================
+
 # Use non-root user
 USER appuser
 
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -81,8 +82,20 @@ ARG REACT_APP_GTM_PREVIEW
 # Build application
 RUN yarn build
 
+# ==========================================
+FROM helsinkitest/node:12-slim as production
+# ==========================================
+# Use non-root user
+USER appuser
+
+COPY --chown=appuser:appuser package.prod.json /app/package.json
```
jorilindell, line 91: You could also copy yarn.lock or use specific canvas version to always use right canvas version

docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
     image: events-helsinki-client
     build:
       context: .
-      target: ${DOCKER_TARGET:-development}
+      target: ${DOCKER_TARGET:-production}
```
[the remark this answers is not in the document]
Sartonon: This should development by default. Forgot to change it back :)

Dockerfile
Sartonon: I thought I removed "^" from the package.prod.json. But it seems to be there. Do you think that is enough to have the correct version? Would copying yarn.lock only install canvas or all the dependencies?

jorilindell: Removing ^ should be enough. Good question about yarn.lock. I think no, but to be sure I have to test it.

jorilindell: Copying yarn.lock wan't good idea :D 
Screenshot 2020-12-19 at 14 33 54

jorilindell (approved):
This is really nice improvement. Looks good to me

PR #167, opened 2020-12-22
Fix: Lint stage skips ignored files

nenonja:
## Description :sparkles:

If you have regenerated `graphql.tsx` which is ignored, you get the following failure on pre-commit becayse the file is picked up by changed files:
 `0:0 warning File ignored because of a matching ignore pattern. Use “--no-ignore” to override`
 
This will fix the problem.

### DIFF
--- a/.lintstagedrc.js
+++ b/.lintstagedrc.js
@@ -2,6 +2,14 @@ const { CLIEngine } = require('eslint');
 
 const cli = new CLIEngine({});
 
+/**
+ * ESLint throws out warning:
+ * "File ignored because of a matching ignore pattern. Use "--no-ignore" to override"
+ * that breaks the linting process when --max-warnings=0 is set.
+ * Based on the discussion from this issue: https://github.com/eslint/eslint/issues/9977,
+ * it seems that using the outlined script is the best route to fix this: https://github.com/eslint/eslint/issues/9977#issuecomment-406420893
+ * See also: https://openbase.io/js/lint-staged/documentation
+ */
 module.exports = {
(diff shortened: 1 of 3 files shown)


.lintstagedrc.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = {
+  '**/*.{ts,tsx,js,jsx}': [
+    (changedFiles) =>
+      'yarn format-code ' +
+      changedFiles.filter((file) => !cli.isPathIgnored(file)).join(' '),
+    'yarn test:staged',
+    'prettier --write',
+  ],
+};
```
Sartonon, line 13: I think it would be nice to have link to the solution in the docs and explanation why we need this

nenonja: To the readme or just add commenting block to this file explaining situation?

.lintstagedrc.js
```diff
@@ -0,0 +1,21 @@
...
+const cli = new CLIEngine({});
+
+/**
+ * ESLint throws out warning:
+ * "File ignored because of a matching ignore pattern. Use "--no-ignore" to override"
+ * that breaks the linting process when --max-warnings=0 is set.
+ * Based on the discussion from this issue: https://github.com/eslint/eslint/issues/9977,
+ * it seems that using the outlined script is the best route to fix this: https://github.com/eslint/eslint/issues/9977#issuecomment-406420893
+ * See also: https://openbase.io/js/lint-staged/documentation
+ */
```
Sartonon, line 12: this is good :)

Sartonon (approved):

PR #169, opened 2020-12-22
TH-899: Add isFree checkbox and search logic to courses search

melniiv:

### DIFF
--- a/src/domain/courseSearch/Search.tsx
+++ b/src/domain/courseSearch/Search.tsx
@@ -63,13 +63,14 @@ const Search: React.FC<Props> = () => {
   const [isCustomDate, setIsCustomDate] = React.useState<boolean>(false);
   const divisionOptions = useDivisionOptions();
 
-  const { alsoOngoingCourses } = getSearchFilters(searchParams);
+  const { alsoOngoingCourses, isFree } = getSearchFilters(searchParams);
 
   const searchFilters = {
     alsoOngoingCourses,
     categories: selectedCategories,
     dateTypes: selectedDateTypes,
     divisions: selectedDivisions,
+    isFree,
     places: selectedPlaces,
     text: selectedTexts,
     start,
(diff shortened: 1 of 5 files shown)


src/domain/courseSearch/Search.tsx
```diff
@@ -93,6 +96,7 @@ const Search: React.FC<Props> = () => {
     setSelectedPlaces(places);
     setSelectedTexts(text);
     setSelectedDateTypes(dateTypes);
+    setIsFree(isFree || false);
```
Sartonon, line 99: This style caught my eye, nothing wrong with this implementation but

`setIsFree(!!isFree)` or `setIsFree(Boolean(isFree))` could also be used.

No need to change this, use the one that you think is best.

Also noticed that this is done differently in eventSearch/Search.tsx. Check that one out, do we need the state at all?

Sartonon (commented):
Consider if we should use the same implementation here as in eventSearch/Search.tsx

src/domain/courseSearch/Search.tsx
melniiv: yea, the implementations is a bit different in those 2. in courses its inside the hook, but not in events.. even though its the same functionality. i just didnt try to refactor, as it might be due to some business logic diffs which i dont see now.

Sartonon: Ok, I will check when I'm back at computer

melniiv: sure

nenonja:
Looks good!

src/domain/courseSearch/Search.tsx
Sartonon: Compare isFree to alsoOngoingCourses in this file. isFree could be handled the same way?

Sartonon (approved):
Looks good! 👍 🥳

PR #170, opened 2020-12-22
Th-908: Bottom banner

nenonja:
## Description :sparkles:
Add bottom banner to front page. Also refactored topBanner to use Landingpage > topBanner fields.

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-908
## Testing :alembic:
start server locally and:
- Change bottom banner from cms: https://cms.test.kuva.hel.ninja/admin/pages/60/edit/ see how it changes on front page:  http://localhost:3000/fi/home 
- Try to change some other page: https://cms.test.kuva.hel.ninja/admin/pages/81/edit/ and preview the change:
http://localhost:3000/fi/home/81


src/domain/banner/bannerConstants.ts
```diff
@@ -0,0 +1 @@
+export const BANNER_SOME_IMAGE = '/images/activities_SoMe-share.jpg';
```
Sartonon, line 1: naming of this file is not in line with other files with constant values. Shoul we also name this "constants.ts"?

src/domain/landingPage/LandingPagePreview.tsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
```
Sartonon, line 1: is there need to change this import style?

src/domain/banner/bannerUtils.ts
```diff
@@ -0,0 +1,110 @@
+import { BannerPage, Maybe } from '../../generated/graphql';
```
Sartonon, line 1: naming of the file is not in line with the other constant files. Not saying this worse, might be better actually but should we change the naming of the files in all places?

src/domain/landingPage/__tests__/LandingPagePreview.test.tsx
```diff
@@ -46,24 +56,35 @@ const renderComponent = () =>
...
     expect(
-      screen.getByRole('heading', { name: landingPageTitle })
+      screen.getByRole('heading', { name: topBannerTitle })
+    ).toBeInTheDocument();
+  });
+  await waitFor(() => {
+    expect(
+      screen.getByRole('heading', { name: bottomBannerTitle })
     ).toBeInTheDocument();
   });
```
Sartonon, line 89: `await screen.findByRole('heading', { name: bottomBannerTitle })`

would be better if we are waiting for an element

Sartonon: that avoids use of waitFor

src/util/mockDataUtils.ts
```diff
@@ -241,28 +242,45 @@ export const fakeLandingPages = (
...
+    heroTopLayerImage: fakeLocalizedCmsImage(),
+    keywords: fakeLocalizedCmsKeywords(),
+    metaInformation: fakeLocalizedObject(),
+    pageTitle: fakeLocalizedObject(),
+    socialMediaImage: fakeLocalizedCmsImage(),
+    title: fakeLocalizedObject(),
+    titleAndDescriptionColor: fakeLocalizedObject('BLACK'),
+    __typename: 'LandingPage',
+    ...overrides,
+  };
```
Sartonon, line 264: lets not get rid of merge function here. it merges objects deeploy and that is why it was used here

src/util/mockDataUtils.ts
```diff
@@ -241,28 +242,45 @@ export const fakeLandingPages = (
...
+};
 
+export const fakeBanner = (overrides?: Partial<BannerPage>): BannerPage => {
+  return {
```
Sartonon, line 268: here we could add merge

src/domain/banner/bannerUtils.ts
nenonja: I would rather rename them all than have multiple files with the same name. At least in intelliJ it makes searching by filenames easier

src/util/mockDataUtils.ts
nenonja: mixed it up with lodash assign which can be replaced by es6 spread.

src/domain/banner/bannerUtils.ts
Sartonon: vscode nicely shows the path for all the found files so having same names is not necessarily problem there but the case might be different for IntelliJ
Screenshot 2020-12-23 at 8 44 00

Sartonon: Screenshot 2020-12-23 at 8 46 39

another example

Sartonon: Screenshot 2020-12-23 at 8 47 27

and when file is open you can see the path

nenonja: IntelliJ have the same features but the problem is perhaps when you want to find all LandingPage related file:
Screenshot 2020-12-23 at 10 56 59 AM
You wont find the utils.ts
But now when you search banner releated file, bannerUtils is visible:
Screenshot 2020-12-23 at 10 57 13 AM

src/domain/landingPage/LandingPagePreview.tsx
nenonja: Typescripts complains about it on intelliJ, I think that's more proper way to import.

src/domain/landingPage/__tests__/LandingPagePreview.test.tsx
nenonja: fixed

src/util/mockDataUtils.ts
nenonja: now fixed

src/util/mockDataUtils.ts
nenonja: fixed

src/domain/landingPage/LandingPagePreview.tsx
Sartonon: in tsconfig we have

`"esModuleInterop": true`

so we can import "normally"

Sartonon (approved):
Looks good!

PR #174, opened 2020-12-29
TH-901: Add course categories with keyword_set

melniiv:
Adds the internal logic for handling course category keyword_sets.
The actual set ids not implemented. Waiting for backend.


src/domain/eventSearch/Search.tsx
```diff
@@ -60,6 +64,10 @@ const Search: React.FC<Props> = ({ scrollToResultList }) => {
   const [selectedTexts, setSelectedTexts] = React.useState<string[]>([]);
   const [autosuggestInput, setAutosuggestInput] = React.useState('');
 
+  /*const { data: landingPageData } = useCourseCategoriesListQuery({
+    variables: { visibleOnFrontpage: true },
+  });*/
+
```
Sartonon, line 70: what is this? 🤔

src/domain/eventSearch/constants.ts
```diff
@@ -100,18 +122,33 @@ export const SPORT_KEYWORDS = [
...
+};
+
+export const MAPPED_CATEGORIES: Record<string, Record<string, string>> = {
+  [String(EVENT_SEARCH_SOURCES.COURSES)]: MAPPED_COURSE_CATEGORIES,
+  [String(EVENT_SEARCH_SOURCES.EVENTS)]: MAPPED_EVENT_CATEGORIES,
+};
+
+export const MAPPED_CATEGORY_TERMS: Record<string, string> = {
+  [String(EVENT_SEARCH_SOURCES.COURSES)]: 'keyword_set_AND',
+  [String(EVENT_SEARCH_SOURCES.EVENTS)]: 'keyword',
```
Sartonon, line 151: is String() function redundant here? Nothing seems to break when I remove it

src/domain/eventSearch/utils.tsx
```diff
@@ -215,10 +228,14 @@ export const getEventSearchVariables = ({
...
   }
 
+  const keywords =
+    MAPPED_CATEGORY_TERMS[searchSource || EVENT_SEARCH_SOURCES.EVENTS];
   const mappedCategories: string[] = categories
-    .map((category) => MAPPED_CATEGORIES[category])
+    .map(
+      (category) =>
+        MAPPED_CATEGORIES[searchSource || EVENT_SEARCH_SOURCES.EVENTS][category]
+    )
```
Sartonon, line 237: at line 179 you could add default value to searchSource:

`searchSource = EVENT_SEARCH_SOURCES.EVENTS,`

Sartonon: Then there would be no need for || which makes this little bit hard to read

src/domain/eventSearch/utils.tsx
```diff
@@ -229,7 +246,7 @@ export const getEventSearchVariables = ({
     end,
     include,
     isFree: isFree || undefined,
-    keyword: [...(keyword ? keyword : []), ...mappedCategories],
+    [keywords]: [...(keyword ? keyword : []), ...mappedCategories],
```
Sartonon, line 249: I think that this naming for the keywords little bit confusing. Naming indicates that it contains keywords but is actually just name for the parameter.

Maybe something like keywordsParamName woul be better? Or maybe you can come up with event better name :)

Sartonon (commented):
Looks pretty good to me at this point :) Couple improvement ideas.

src/domain/courseSearch/__tests__/CourseSearchPageContainer.test.tsx
```diff
@@ -369,3 +370,8 @@ it('should show toastr message when loading next event page fails', async () =>
 
   expect(toast.error).toBeCalledWith(translations.eventSearch.errorLoadMode);
 });
+*/
+
+it('renders title and search fields', async () => {
+  expect(true).toBe(true);
+});
```
Sartonon, line 377: Should this have TODO?

src/domain/eventSearch/Search.tsx
melniiv: ah, i forgot to delete it

src/domain/eventSearch/constants.ts
melniiv: actually without it typescript gives error

src/domain/eventSearch/utils.tsx
melniiv: it depends if we use eent source or they will create the keyword set in courses. i would still want to see api

src/domain/eventSearch/constants.ts
Sartonon: hmm has it something to do with typescript version? for me there we no errors

src/domain/eventSearch/utils.tsx
melniiv: yea, i was thinking to ask about some name ideas, if we useParamName as a postfix, then it would be nice to make it a convention to use this naming for var params.

src/domain/courseSearch/__tests__/CourseSearchPageContainer.test.tsx
melniiv: yes.

src/domain/eventSearch/utils.tsx
Sartonon: But you are using `EVENT_SEARCH_SOURCES.EVENTS` as a default value anyway atm for `MAPPED_CATEGORY_TERMS` and `MAPPED_CATEGORIES`. So for now at least think it would make more sense to have the default value set to searchSource. If i'm understading this correctly

Sartonon: of course thing can change and event search also needs to use keywordSet but then this has to be thought again. Or if you have some idea how we could handle this now to have it work with keywordSet from event search also

melniiv: yes, i am using it as default as its optional param, and undefined cannot be used as index. this method is used a lot by event search, so when no search source, then we use EVENT_SEARCH_SOURCES.EVENTS (and as default), when we define search source, then we use different. Now i defined 2 sources, events and courses, which means basically that we query linked_events or linked_courses. I still think that they will create the keyword_set for courses under the linked_courses, if not, i can fix that later when implementing the rest.

src/domain/eventSearch/constants.ts
melniiv: hmm, you are right. now its not giving error, maybe during development i had mistake somewhere else

Sartonon (approved):

PR #175, opened 2020-12-29
th-914: add tests to GitHub actions

nenonja:
## Description :sparkles:
Add github actions for lint, unit/component tests and testcafe tests. 
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-914
## Testing :alembic:
if all checks are passes this shouldbe good
Needed to use macos-latest image for component tests as the canvas measureText seems to be os-dependent: https://stackoverflow.com/questions/46653569/canvas-measuretext-differences-on-browsers-are-huge


package.json
```diff
@@ -79,16 +79,17 @@
...
     "build": "yarn build:client && yarn build:server",
     "build:client": "react-scripts build",
     "build:server": "NODE_ENV=production webpack --progress --mode production --config config/webpack.server.config.js",
     "codegen": "graphql-codegen -r dotenv/config dotenv_config_path=.env.development.local  --config codegen.yml",
-    "test": "react-scripts test --env=jest-environment-jsdom-sixteen",
-    "test:staged": "react-scripts test --env=jest-environment-jsdom-sixteen --watchAll=false --findRelatedTests",
-    "test:coverage": "react-scripts test --coverage --env=jest-environment-jsdom-sixteen --watchAll=false",
-    "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
+    "test": "TZ=Europe/Helsinki react-scripts test --env=jest-environment-jsdom-sixteen",
+    "ci": "CI=true yarn test --verbose --runInBand --coverage",
```
Sartonon, line 89: is runInBand more performant in github action?

package.json
```diff
@@ -79,16 +79,17 @@
...
+    "test:staged": "TZ=Europe/Helsinki react-scripts test --env=jest-environment-jsdom-sixteen --watchAll=false --findRelatedTests",
+    "test:coverage": "TZ=Europe/Helsinki react-scripts test --coverage --env=jest-environment-jsdom-sixteen --watchAll=false",
+    "test:debug": "TZ=Europe/Helsinki react-scripts --inspect-brk test --runInBand --no-cache",
```
Sartonon, line 92: noticed now that these could use the "test" script as base? same as in "ci" script

package.json
```diff
@@ -79,16 +79,17 @@
     "xml-js": "^1.6.11"
   },
   "scripts": {
-    "start": "react-scripts start",
-    "start:server": "node build/server.js",
+    "start": "TZ=Europe/Helsinki react-scripts start",
+    "start:server": "TZ=Europe/Helsinki node build/server.js",
```
Sartonon, line 83: are these env variables required here?

Sartonon:
Screenshot 2020-12-30 at 8 21 29
Should codecov show something?

Sartonon:
Hmm so is it bercause we don't have codecoverage commit fro master branch yet?

.github/workflows/browser-tests.yml
```diff
@@ -0,0 +1,48 @@
...
+    - cron: '0 7 * * 1-5'
+  workflow_dispatch:
+
+jobs:
+  test:
+    name: Run browser tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
Sartonon, line 18: the other project had windows here also? how does it work if we would also have it?

package.json
nenonja: While our docker host has UTC-timezone, this seems to be only way to set the timezone for the node server: https://stackoverflow.com/questions/8083410/how-can-i-set-the-default-timezone-in-node-js
Other question is that do we need that anyway, since server is only used for ssr?

package.json
nenonja: Not sure, this was just used in venepaikka script (for ci only) and counted on that.

package.json
Sartonon: I think this should not be needed

nenonja:
> Hmm so is it bercause we don't have codecoverage commit fro master branch yet?

Yes I think it's the reason

.github/workflows/browser-tests.yml
nenonja: Let's try. Perhaps it's good idea in that sense that most end users have windows chrome.

package.json
nenonja: fixed

package.json
nenonja: This can be a problem at some point if we format or parse timestamps on ssr: https://stackoverflow.com/questions/24655522/should-the-server-or-the-client-handle-timezone-when-sending-receiving-dates
One option is to use only epoch time (milliseconds from 1970) and then render format on client side on user's timezone.
This might be a good idea especially with internet events where user can join anywhere in the world. On the other hand many users might have strong assumption that physical event times are shown in helsinki timezone time.
I can remove this as it is not a problem for now.

package.json
nenonja: no difference. Removed runInBand

.github/workflows/browser-tests.yml
nenonja: windows-latest worked

Sartonon (approved):

PR #181, opened 2021-01-08
Switch to master branch in Gitlab Staging pipeline

quyenlq:


nenonja (approved):
Looks good. Talked with devops team and yes we should not remove anything from gitlab conf for now.

PR #184, opened 2021-01-12
TH-915: Add event search acceptance tests

nenonja:
## Description :sparkles:
- event search acceptance tests
- use page component model for selectors, actions and expectations.
- run tests hourly on master
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-915


browser-tests/event-search/eventSearchPage.testcafe.ts
```diff
@@ -0,0 +1,48 @@
...
+    await t
+      .expect(
+        screen.findByRole('checkbox', { name: neighbourhood.name.fi }).exists
+      )
+      .ok();
+  }
+});
+/**
+ * Note: This is kinda slow test.
+ * Perhaps we could check just some of the places instead of all of them
```
Sartonon, line 35: I think testing this with like 2 would be enough

nenonja: It selects now 3 random places to check

src/domain/event/eventCard/EventCard.tsx
```diff
@@ -85,79 +85,85 @@ const EventCard: React.FC<Props> = ({ event }) => {
...
-            {!!startTime &&
-              getDateRangeStr({
-                start: startTime,
-                end: endTime,
-                locale,
-                includeWeekday: false,
-                includeTime: true,
-                timeAbbreviation: t('commons.timeAbbreviation'),
-              })}
+    <span data-testid={event.id}>
```
Sartonon, line 88: span doesn't seem right to me here? Should we use div or add testid to Link

Sartonon:
Biggest issue I see here is that some changed made to the code are gonna introduce some VERY nasty merge conflicts with events-helsinki-2 branch. Could this be just added to that branch and merge it to master later when that is ready?

Otherwise we should test if we can easily resolve those conlficts before mergin..

Anyway I will continue tomorrow with this review.

nenonja:
> Biggest issue I see here is that some changed made to the code are gonna introduce some VERY nasty merge conflicts with events-helsinki-2 branch. Could this be just added to that branch and merge it to master later when that is ready?
> 
> Otherwise we should test if we can easily resolve those conlficts before mergin..
> 
> Anyway I will continue tomorrow with this review.

I think that there is not much changes to actual code (some data-testid:s added). There will be conflicts on testcafe test side but I can take responsibility on them. Most likely I will just override all changes under browser-tests folder with these changes and then fix the tests if something fails. Most likely observer js error will occur again but there will be a better solution for that which can be put on the testcafe configs.

src/domain/event/eventCard/EventCard.tsx
nenonja: I was able to remove it and put it to Link

Sartonon:
Now that spans are deleted, merge conflicts won't be so bad

browser-tests/event-search/eventSearchPage.components.ts
```diff
@@ -0,0 +1,279 @@
...
+        );
+      },
+    };
+    const expectations = {
+      async isPresent() {
+        await t.expect(selectors.component().exists).ok(getErrorMessage(t));
+      },
+      async neighborhoodOptionIsPresent(neighborhood: Neighborhood) {
+        await this.isPresent();
+        t.ctx.neighborhood = neighborhood;
```
Sartonon, line 71: I'm not quite sure where these are used? Are some of them redundant or just added in case they are used later?

Sartonon: Oh ok they are used in the error message :)

browser-tests/event-search/eventSearchPage.testcafe.ts
```diff
@@ -0,0 +1,142 @@
...
+    await components.eventCard(event).expectations.isPresent();
+    await components
+      .eventCard(event)
+      .expectations.containsDate(getDate(dateRange));
+    await searchContainer.actions.openDateFilters();
+    await searchContainer.actions.selectDateRange(dateRange); // unselect previous choice
+  }
+});
+
+const getDate = (dateRange: string) => {
```
Sartonon, line 129: Does this belong to utils folder?

browser-tests/event-search/eventSearchPage.components.ts
```diff
@@ -0,0 +1,279 @@
...
+        return this.withinComponent().findByRole('checkbox', {
+          name: place.name.fi,
+        });
+      },
+      searchInput() {
+        return this.withinComponent().findByPlaceholderText(
+          'Kirjoita hakusana, esim. rock tai jooga'
+        );
+      },
+    };
```
Sartonon, line 64: To me seems like these don't need to be functions that are called every time you need the element. These could be just functions and that are not nested in selectors object (then we could also avoid using this keyword). "This" keyword forces us to use this api in a specific way to work correctly. These functions can be then wrapped with selectors property in return statemen

Sartonon: Like this:

```
    const component = screen.findByTestId('searchContainer');
    const withinComponent = within(screen.getByTestId('searchContainer'));
    const searchButton = withinComponent.findByRole('button', { name: 'Hae' });
    const dateRangeFilter = withinComponent.findByLabelText(
      'Valitse ajankohta'
    );
    const neighborhoodFilter = withinComponent.findByLabelText('Etsi alue');
    const placeFilter = withinComponent.findByLabelText('Etsi tapahtumapaikka');
    const placeSearchInput = withinComponent.findByLabelText(
      'Kirjoita hakusana'
    );
    const searchInput = withinComponent.findByPlaceholderText(
      'Kirjoita hakusana, esim. rock tai jooga'
    );
    const neighborhoodCheckbox = (n: Neighborhood) =>
      withinComponent.findByRole('checkbox', {
        name: n.name.fi,
      });
    const dateRangeCheckbox = (range: string) =>
      withinComponent.findByRole('checkbox', {
        name: range,
      });
    const placeCheckbox = (place: PlaceFieldsFragment) =>
      withinComponent.findByRole('checkbox', {
        name: place.name.fi,
      });
```

Sartonon (commented):
Tests are looking very good. Some very nice abstractions that makes the test files very easy to read. In the helper functions I would still think if they could be named better. 

For examples then I see `expectations.isPresent();` I'm not sure what should be present and I have to check some other place to remember the context. Also `component()` could be `headerComponent()`

Also the queries don't need to be functions that are called every single time. They can be called once and saved to a const, then they can be used without calling them.

```
    const component = screen.findByTestId('searchContainer');
    const withinComponent = within(screen.getByTestId('searchContainer'));
    const searchButton = withinComponent.findByRole('button', { name: 'Hae' });
    const dateRangeFilter = withinComponent.findByLabelText(
      'Valitse ajankohta'
    );
    const neighborhoodFilter = withinComponent.findByLabelText('Etsi alue');
    const placeFilter = withinComponent.findByLabelText('Etsi tapahtumapaikka');
    const placeSearchInput = withinComponent.findByLabelText(
      'Kirjoita hakusana'
    );
    const searchInput = withinComponent.findByPlaceholderText(
      'Kirjoita hakusana, esim. rock tai jooga'
    );
    const neighborhoodCheckbox = (n: Neighborhood) =>
      withinComponent.findByRole('checkbox', {
        name: n.name.fi,
      });
    const dateRangeCheckbox = (range: string) =>
      withinComponent.findByRole('checkbox', {
        name: range,
      });
    const placeCheckbox = (place: PlaceFieldsFragment) =>
      withinComponent.findByRole('checkbox', {
        name: place.name.fi,
      });

    const searchComponentIsPresent = async () => {
      await t.expect(component.exists).ok(getErrorMessage(t));
    };
    const neighborhoodOptionIsPresent = async (neighborhood: Neighborhood) => {
      await searchComponentIsPresent();
      t.ctx.neighborhood = neighborhood;
      await t
        .expect(neighborhoodCheckbox(neighborhood).exists)
        .ok(getErrorMessage(t));
    };
```

nenonja:
> withinComponent

First I tried to do this without functions but it did not work. The problem is that the screen is behind all selectors, and it changes over time. So the selector must be evaluated at the moment of call.

I can try to rename isPresent and Component. I think that the real improvement would be to use classes + inheritance to avoid some copy-paste but now it's purely functional model. testcafe recommends to use classes: https://devexpress.github.io/testcafe/documentation/guides/concepts/page-model.html
I found this link after I had already done my own model :)

nenonja:
> Now that spans are deleted, merge conflicts won't be so bad

Alas, the selector does not work anymore. I investigate it...

browser-tests/event-search/eventSearchPage.components.ts
nenonja: The default error message wont tell much. However now it does not print html anymore, I still have to improve the error message

nenonja: Now it prints always the "within" component:
https://github.com/City-of-Helsinki/events-helsinki-ui/pull/184/files#diff-70f4d2d9a0ec9b325b6d84d4c1601f1dfb7e8cf3d0dfb68b93d6c54107cc66c1R4-R22

browser-tests/event-search/eventSearchPage.testcafe.ts
nenonja: moved

nenonja:
> > Now that spans are deleted, merge conflicts won't be so bad
> 
> Alas, the selector does not work anymore. I investigate it...

fixed

browser-tests/event-search/eventSearchPage.components.ts
nenonja: here is comment below why it does not work

Sartonon:
> > Also the queries don't need to be functions that are called every single time. They can be called once and saved to a const, > then they can be used without calling them. Also we can avoid this keywords like this:
> 
> First I tried to do this without functions but it did not work. The problem is that the screen is behind all selectors, and it changes over time. So the selector must be evaluated at the moment of call.
> 
> > For examples then I see expectations.isPresent(); I'm not sure what should be present and I have to check some other place > to remember the context. Also component() could be headerComponent()
> 
> I can try to rename isPresent and Component. I think that the real improvement would be to use classes + inheritance to avoid some copy-paste but now it's purely functional model. Testcafe recommends to use classes: https://devexpress.github.io/testcafe/documentation/guides/concepts/page-model.html
> It uses`import { Selector, t } from 'testcafe';` but I was not able to make `t` work like this. Perhaps beause of testcafe-testing-library. So it must be passed to all constructors, and the resolution might not be as clean as it is in the testcafe example.

When I tested without calling functions every time and it worked, was I lucky?

Sartonon:
https://github.com/testing-library/testcafe-testing-library/blob/master/src/index.ts

There is:

```
const bindFunction = <
  T extends QueryName,
  Options = Parameters<typeof queries[T]>[2]
>(
  queryName: T
) => {
  const query = queryName.replace("find", "query") as T;
  return (matcher: Matcher, options?: Options) => {
    return Selector(
      () =>
        window.TestingLibraryDom[query](document.body, matcher, options) as
          | Node
          | Node[]
          | NodeList
          | HTMLCollection,
      {
        dependencies: { query, matcher, options },
      }
    );
  };
}; 
```

To me looks like the queries just return Selector with should update without calling it again. But I'm not 100% sure how Selector works

nenonja:
> https://github.com/testing-library/testcafe-testing-library/blob/master/src/index.ts
> 
> There is:
> 
> ```
> const bindFunction = <
>   T extends QueryName,
>   Options = Parameters<typeof queries[T]>[2]
> >(
>   queryName: T
> ) => {
>   const query = queryName.replace("find", "query") as T;
>   return (matcher: Matcher, options?: Options) => {
>     return Selector(
>       () =>
>         window.TestingLibraryDom[query](document.body, matcher, options) as
>           | Node
>           | Node[]
>           | NodeList
>           | HTMLCollection,
>       {
>         dependencies: { query, matcher, options },
>       }
>     );
>   };
> }; 
> ```
> 
> To me looks like the queries just return Selector with should update without calling it again. But I'm not 100% sure how Selector works

It could work, the component model has developed since I last time tried, and I think previous version the component was initialized in before-test clause which might have affected to it. I give it a second try. Also I have temptation to refactor this model to class-based so that all common things would be in abstract class and pageComponents are inherited from there. This enforces to implement necessary parts when we going to have more pagecomponents for new tests.

nenonja:
> > https://github.com/testing-library/testcafe-testing-library/blob/master/src/index.ts
> > There is:
> > ```
> > const bindFunction = <
> >   T extends QueryName,
> >   Options = Parameters<typeof queries[T]>[2]
> > >(
> >   queryName: T
> > ) => {
> >   const query = queryName.replace("find", "query") as T;
> >   return (matcher: Matcher, options?: Options) => {
> >     return Selector(
> >       () =>
> >         window.TestingLibraryDom[query](document.body, matcher, options) as
> >           | Node
> >           | Node[]
> >           | NodeList
> >           | HTMLCollection,
> >       {
> >         dependencies: { query, matcher, options },
> >       }
> >     );
> >   };
> > }; 
> > ```
> > 
> > 
> > To me looks like the queries just return Selector with should update without calling it again. But I'm not 100% sure how Selector works
> 
> It could work, the component model has developed since I last time tried, and I think previous version the component was initialized in before-test clause which might have affected to it. I give it a second try. Also I have temptation to refactor this model to class-based so that all common things would be in abstract class and pageComponents are inherited from there. This enforces to implement necessary parts when we going to have more pagecomponents for new tests.

I looked into it and I still think the functions have benefits and does not clutter code too much, for example:
`    const selectors = {
      searchBanner() {
        return screen.findByTestId('searchContainer');
      },
      withinSearchBanner() {
        t.ctx.withinTestId = 'searchContainer';
        return within(screen.getByTestId(t.ctx.withinTestId));
      },
      searchButton() {
        return this.withinSearchBanner().findByRole('button', { name: 'Hae' });
      },
      dateRangeFilter() {
        return this.withinSearchBanner().findByLabelText('Valitse ajankohta');
      },`
      
      sets `withinTestId` everytime when within-selector is used which in turn is used in error message if selector fails, we get html printed inside within selector instead of whole html which is default on testcafe (which is even cutted becuase html is too long and does not debuging help at all).
      Here is example when build fails how error message is now detailed and helpful: https://github.com/City-of-Helsinki/events-helsinki-ui/runs/1760714956?check_suite_focus=true

Sartonon:
Having the functions is not a big problem for me but the usage of this keyword here is little bit bothering. This would fail if you decided to call the function without pointin to selector object:  `selector.something()` -> `something()`. This could happen if you would for example deconstruct the selectors

browser-tests/event-search/eventSearchPage.components.ts
```diff
@@ -0,0 +1,274 @@
...
+      },
+      searchButton() {
+        return this.withinSearchBanner().findByRole('button', { name: 'Hae' });
+      },
+      dateRangeFilter() {
+        return this.withinSearchBanner().findByLabelText('Valitse ajankohta');
+      },
+      dateRangeCheckbox(range: string) {
+        t.ctx.findByRole = ['checkbox', { name: range }];
+        return this.withinSearchBanner().findByRole.apply(
```
Sartonon, line 38: Do we have some issue if we don't use .apply?

nenonja: Idea was to make shorter code, another option is:
```
        t.ctx.findByRole = { 
          matcher: 'checkbox', 
          options: { name: range } 
        };
        return this.withinSearchBanner().findByRole(
          t.ctx.findByRole.matcher,
          t.ctx.findByRole.options
        );
```
Good thing about ` t.ctx.findByRole ` is that it will tell exactly in error message what we tried to find if selector fails. testcafe's default error message does not tell the searched value `range` anywhere.

nenonja:
> Having the functions is not a big problem for me but the usage of this keyword here is little bit bothering. This would fail if you decided to call the function without pointin to selector object: `selector.something()` -> `something()`. This could happen if you would for example deconstruct the selectors

Forgot about destructor, I can fix "this".

nenonja:
> > Having the functions is not a big problem for me but the usage of this keyword here is little bit bothering. This would fail if you decided to call the function without pointin to selector object: `selector.something()` -> `something()`. This could happen if you would for example deconstruct the selectors
> 
> Forgot about destructor, I can fix "this".

fixed

browser-tests/event-search/eventSearchPage.components.ts
Sartonon: Oh now I understand. Didn't rmember exact details how apply works, but I can see the benefit when now

Sartonon:
The way we are setting context for errors now in the functions seems pretty big manual work. We could implement function that returns those query functions and does all that ctx stuff there?

nenonja:
> The way we are setting context for errors now in the functions seems pretty big manual work. We could implement function that returns those query functions and does all that ctx stuff there?

I rethink this and resolution was to inject some code to the screen and within functions itself, now the code is clean and error context is set "behind the scenes". Next improve would be to use class implementation that would enforce to create components by extending a base class that sets context correctly.

browser-tests/utils/testcafe.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+export const withinContext = (t: TestController): typeof within => (selector) =>
+  injectCallback(within(selector), (key, params) => {
+    t.ctx.latestSearch = {
+      ...(t.ctx.within && { within: t.ctx.within }),
+      key,
+      params: params?.length === 1 ? params[0] : params,
+    };
+  });
+
+const injectCallback = <Key extends string, Obj extends Record<Key, unknown>>(
```
Sartonon, line 24: `<Key extends string, Obj extends Record<Key, unknown>>`

Not sure why that works but:

`<Obj extends Record<Key, unknown>>` doesn't

at least this doesn't give more type safety

browser-tests/utils/testcafe.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  });
+
+export const withinContext = (t: TestController): typeof within => (selector) =>
+  injectCallback(within(selector), (key, params) => {
+    t.ctx.latestSearch = {
+      ...(t.ctx.within && { within: t.ctx.within }),
+      key,
+      params: params?.length === 1 ? params[0] : params,
+    };
+  });
```
Sartonon, line 22: this is good, good job :)

Sartonon (approved):
Looking very good now, great job!

browser-tests/utils/testcafe.utils.ts
nenonja: fixed

PR #186, opened 2021-01-14
Error logging

CodingSam:
## Description :sparkles:
Add error logging for production builds

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -122,9 +122,11 @@ app.use(async (req: Request, res: Response) => {
     // await getDataFromTree(el) fails if any GraphQl request fails.
     // Try to generate html withour apollo state
     // TODO: Downside is that page is re-rendered on client which causes blinking
+    console.error(e);
     try {
       generateHtmlAndSendResponse();
     } catch (e) {
+      console.error(e);
       res.send('Something went very very wrong. Failed to catch error.');
     }
   }
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -116,19 +116,15 @@ app.use(async (req: Request, res: Response) => {
   try {
     // Executes all graphql queries for the current state of application
     await getDataFromTree(el);
+  } catch (e) {
+    console.error('Error - GrapQl requests failed', e);
+  }
 
+  try {
     generateHtmlAndSendResponse();
   } catch (e) {
-    // await getDataFromTree(el) fails if any GraphQl request fails.
-    // Try to generate html withour apollo state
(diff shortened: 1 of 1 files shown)


src/server/index.ts
```diff
@@ -122,9 +122,11 @@ app.use(async (req: Request, res: Response) => {
     // await getDataFromTree(el) fails if any GraphQl request fails.
     // Try to generate html withour apollo state
     // TODO: Downside is that page is re-rendered on client which causes blinking
+    console.error(e);
```
Sartonon, line 125: Hmm, would it be beneficial if we added some context to the error? Some string that says where things went wrong? When this error happens?

I mean this might be enough, just wondering

CodingSam: sure, I can add that :+1:

nenonja:
For failed test, it is unrelated but it seems that sometimes random fakename generator generates same name twice (happened to me too). It could be fixed so that we collect all already generated names and check generated name against it or just generate more complex names. However not in this PR, perhaps create a task to fix it.

Sartonon:
> For failed test, it is unrelated but it seems that sometimes random fakename generator generates same name twice (happened to me too). It could be fixed so that we collect all already generated names and check generated name against it or just generate more complex names. However not in this PR, perhaps create a task to fix it.

Fakers seems to have .unique function that might help with this problem.

src/server/index.ts
```diff
@@ -116,17 +116,15 @@ app.use(async (req: Request, res: Response) => {
   try {
     // Executes all graphql queries for the current state of application
     await getDataFromTree(el);
+  } catch (e) {
+    console.error('Error - GrapQl requests failed', e);
+  }
 
+  try {
```
Sartonon, line 123: This try-catch is no logne nested inthe other catch block like it was? Is is accidental?
```
  try {
    // Executes all graphql queries for the current state of application
    await getDataFromTree(el);

    generateHtmlAndSendResponse();
  } catch (e) {
    // await getDataFromTree(el) fails if any GraphQl request fails.
    // Try to generate html withour apollo state
    // TODO: Downside is that page is re-rendered on client which causes blinking
    try {
      generateHtmlAndSendResponse();
    } catch (e) {
      res.send('Something went very very wrong. Failed to catch error.');
    }
  }
```

CodingSam: Nope, just simplified the code. My solution should have the same outcome :)

Sartonon: oh true :)

Sartonon (approved):
Good :)

PR #189, opened 2021-01-19
Fix for duplicate event names

CodingSam:
## Description :sparkles:
Add fix for duplicate event names

### DIFF
--- a/src/util/mockDataUtils.ts
+++ b/src/util/mockDataUtils.ts
@@ -46,11 +46,25 @@ export const fakeEvents = (
 });
 
 export const fakeEvent = (overrides?: Partial<EventDetails>): EventDetails => {
+  const names: string[] = [];
+
+  const uniqueName = (): string => {
+    const name = faker.name.title();
+
+    if (names.includes(name)) {
+      return uniqueName();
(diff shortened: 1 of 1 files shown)


CodingSam:
Faker has not published a version to npm where the unique function is included. So I made a small recursive function.

nenonja (approved):
Looks good!

src/util/mockDataUtils.ts
```diff
@@ -46,11 +46,25 @@ export const fakeEvents = (
...
+
+    return name;
+  };
+
   return merge(
     {
       id: `hel:${faker.random.uuid()}`,
       internalId: faker.random.uuid(),
-      name: fakeLocalizedObject(faker.name.title()),
+      name: fakeLocalizedObject(uniqueName()),
```
Sartonon, line 67: Is there same risk of having duplicates in other tests too?

CodingSam: Hmm, maybe? What if we went like this now, and if in the future if we still get this error, then we could make a utility which could compare all the fields and make sure there are no duplicates.

PR #194, opened 2021-01-22
TH-892 | More fixes based on comments

Sartonon:
## Description :sparkles:

- Focus outline hidden fix on landing page
- Make other event time list item clickable from the whole area
- Fix back to top button not to focus skip link

### DIFF
--- a/src/domain/app/resetFocus/ResetFocus.tsx
+++ b/src/domain/app/resetFocus/ResetFocus.tsx
@@ -19,7 +19,7 @@ const ResetFocus = (): React.ReactElement => {
     }
   }, [pathname]);
 
-  return <div ref={node} tabIndex={-1} id={resetFocusId}></div>;
+  return <div ref={node} tabIndex={-1} id={resetFocusId} />;
 };
 
 export default ResetFocus;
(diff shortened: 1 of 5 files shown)


nenonja (approved):
Approved, only thing I was thinking that when using keyboard (using tab + press enter) to click "back to top" it would work in the old way and show "siirry sisältöön" link, now keyboard user might have as "lost" feeling when going to top without visible focus. But that is really minor thing.

src/domain/app/resetFocus/ResetFocus.tsx
```diff
@@ -16,7 +19,7 @@ const ResetFocus = (): React.ReactElement => {
     }
   }, [pathname]);
 
-  return <div ref={node} tabIndex={-1}></div>;
+  return <div ref={node} tabIndex={-1} id={resetFocusId}></div>;
```
nenonja, line 22: My IDE complains that tag is empty (so it could be `<div ... />`

Sartonon: fixed

Sartonon:
> Approved, only thing I was thinking that when using keyboard (using tab + press enter) to click "back to top" it would work in the old way and show "siirry sisältöön" link, now keyboard user might have as "lost" feeling when going to top without visible focus. But that is really minor thing.

I think hitting tab before skip link makes sense (after back to top). That is how it works when opening the page or changing route.

PR #202, opened 2021-02-03
Hide similar events from eventpage

nenonja:
## Description :sparkles:
See conversation here: https://helsinkicity.slack.com/archives/C013682SYSD/p1612348876019900
Similar events with lots of canceled events makes marketing campaigns look not so good. Thus similar events will be hidden for the time being. It's done behind feature flag, also introduced crude "karvalakki" version of the feature flag functionality. 
Also added tests that check that feature flag works on and off.


src/domain/event/__tests__/EventPageContainer.test.tsx
```diff
@@ -100,3 +106,32 @@ it("should show error info when event doesn't exist", async () => {
...
+    await waitFor(() => {
+      expect(screen.queryByTestId('loading-spinner')).not.toBeInTheDocument();
+    });
+    expect(
+      screen.queryByRole('heading', {
+        name: translations.event.similarEvents.title,
+      })
+    ).toBeInTheDocument();
+  });
+  it('doesnt shor similar events when flag is off', async () => {
```
Sartonon, line 124: typo: shor

src/domain/event/EventPageContainer.tsx
```diff
@@ -53,7 +53,9 @@ const EventPageContainer: React.FC = () => {
               {/* Show event content only if event is open */}
               {!eventClosed && <EventContent event={event} />}
               {/* Hide similar event on SSR to make initial load faster */}
-              {isClient && <SimilarEvents event={event} />}
+              {isClient && getFeatureFlags().SHOW_SIMILAR_EVENTS && (
```
Sartonon, line 56: Minor readability improvement would be to call this function in top of this function and destructure SHOW_SIMILAR_EVENTS. But that is debatable :D

Sartonon (approved):
Looks good. Couple comments but approved.

src/domain/event/__tests__/EventPageContainer.test.tsx
nenonja: fixed

src/domain/event/EventPageContainer.tsx
nenonja: fixed

PR #203, opened 2021-02-04
TH-984: Improves courses search target group dropdown

melniiv:

### DIFF
--- a/src/common/components/rangeDropdown/RangeDropdown.tsx
+++ b/src/common/components/rangeDropdown/RangeDropdown.tsx
@@ -66,6 +66,7 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
 
   //set values with prevalidation, rest of validation on blur
   const handleInputChange = (inputType: RANGE_INPUT, val: string) => {
+    console.log('val', val);
     const allowChange =
       (maxInputEndValue && val.length <= maxInputEndValue.length) ?? true;
     if (allowChange) {
@@ -109,9 +110,11 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
 
     switch (inputType) {
       case RANGE_INPUT.MIN:
+        console.log('MIN', getValidatedValue(val), maxInputValue);
         onChange(getValidatedValue(val), maxInputValue);
         break;
       case RANGE_INPUT.MAX:
+        console.log('MAX', minInputValue, getValidatedValue(val));
         onChange(minInputValue, getValidatedValue(val));
         break;
       case RANGE_INPUT.ALL:
--- a/src/common/components/rangeDropdown/RangeDropdown.tsx
+++ b/src/common/components/rangeDropdown/RangeDropdown.tsx
@@ -66,7 +66,6 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
 
   //set values with prevalidation, rest of validation on blur
   const handleInputChange = (inputType: RANGE_INPUT, val: string) => {
-    console.log('val', val);
     const allowChange =
       (maxInputEndValue && val.length <= maxInputEndValue.length) ?? true;
     if (allowChange) {
@@ -110,11 +109,9 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
 
     switch (inputType) {
       case RANGE_INPUT.MIN:
-        console.log('MIN', getValidatedValue(val), maxInputValue);
         onChange(getValidatedValue(val), maxInputValue);
         break;
       case RANGE_INPUT.MAX:
-        console.log('MAX', minInputValue, getValidatedValue(val));
         onChange(minInputValue, getValidatedValue(val));
         break;
       case RANGE_INPUT.ALL:
--- a/src/common/components/rangeDropdown/RangeDropdown.tsx
+++ b/src/common/components/rangeDropdown/RangeDropdown.tsx
@@ -80,27 +80,26 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
 
   //validate values on blur
   const handleInputBlur = (inputType: RANGE_INPUT, val: string) => {
-    const getValidatedMinValue = (val: string): string => {
+    const getValidatedValue = (val: string): string => {
       let resultValue = val;
       if (Number(val) < Number(minInputStartValue)) {
         resultValue = minInputStartValue;
       } else if (Number(val) > Number(maxInputEndValue)) {
         resultValue = maxInputEndValue;
       }
-      if (val && maxInputValue && Number(val) > Number(maxInputValue)) {
+      if (
+        inputType !== RANGE_INPUT.MAX &&
+        val &&
+        maxInputValue &&
+        Number(val) > Number(maxInputValue)
+      ) {
         resultValue = maxInputValue;
-      }
-      return resultValue;
-    };
-
-    const getValidatedMaxValue = (val: string): string => {
-      let resultValue = val;
-      if (Number(val) > Number(maxInputEndValue)) {
-        resultValue = maxInputEndValue;
-      } else if (Number(val) < Number(minInputStartValue)) {
-        resultValue = minInputStartValue;
-      }
-      if (val && minInputValue && Number(val) < Number(minInputValue)) {
+      } else if (
+        inputType !== RANGE_INPUT.MIN &&
+        val &&
+        minInputValue &&
+        Number(val) < Number(minInputValue)
+      ) {
         resultValue = minInputValue;
       }
       return resultValue;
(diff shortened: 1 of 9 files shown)


src/common/components/rangeDropdown/RangeDropdown.tsx
```diff
@@ -64,15 +64,17 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
...
+  //set values with prevalidation, rest of validation on blur
   const handleInputChange = (inputType: RANGE_INPUT, val: string) => {
-    switch (inputType) {
-      case RANGE_INPUT.MIN:
-        onChange(val, maxInputValue);
-        break;
-      case RANGE_INPUT.MAX:
-        onChange(minInputValue, val);
-        break;
+    if (maxInputEndValue ? val.length <= maxInputEndValue.length : true) {
```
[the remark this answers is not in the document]
Sartonon: And could this expression be simplified? :

`maxInputEndValue && val.length <= maxInputEndValue.length ?? true`

Sartonon (commented):
Maybe add some simple test for validation?

src/common/components/rangeDropdown/RangeDropdown.tsx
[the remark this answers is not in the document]
melniiv: fixed

src/domain/courseSearch/Search.tsx
```diff
@@ -311,13 +311,12 @@ const Search: React.FC<Props> = ({ scrollToResultList }) => {
                   minInputFixedValue={'18'}
                   maxInputValue={maxAgeInput}
                   maxInputLabel={t('courseSearch.search.ageLimitMax')}
-                  maxInputEndValue={'100'}
+                  maxInputEndValue={'99'}
```
nenonja, line 314: Just wondering why these values are originally strings and not numbers? Now this causes a lots of converting inside RangeDropDown component

src/common/components/rangeDropdown/RangeDropdown.tsx
```diff
@@ -64,57 +64,60 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
...
+          onChange(minInputValue, val);
+          break;
+      }
     }
   };
 
   //validate values on blur
   const handleInputBlur = (inputType: RANGE_INPUT, val: string) => {
-    const getValidatedMinValue = (val: string): string => {
+    const getValidatedValue = (val: string): string => {
```
nenonja, line 86: This method is quite hard to follow. I think that the previous version of separated `getValidatedMinValue` and `getValidatedMaxValue` was cleaner to read. Also If I understand code could be shorter something like this:
```
 const getValidatedMinValue = (val: string): string => {
return Math.Min(val, maxInputValue , maxInputEndValue)
}
 const getValidatedMaxValue = (val: string): string => {
return Math.Max(minInputStartValue, minInputValue, val)
}
```
It seems that Math functions ignore null values, but behaviour unexpectedly with NaN and undefined, so falsy values should be converted to `null` first: `const normalize(val: string) => Number(val) || null;`
I didnt test this fully so I might be also wrong as well :)

src/common/components/rangeDropdown/RangeDropdown.tsx
```diff
@@ -64,57 +64,60 @@ const RangeDropdown: React.FC<RangeDropdownProps> = ({
...
       let resultValue = val;
       if (Number(val) < Number(minInputStartValue)) {
         resultValue = minInputStartValue;
       } else if (Number(val) > Number(maxInputEndValue)) {
         resultValue = maxInputEndValue;
       }
-      if (val && maxInputValue && Number(val) > Number(maxInputValue)) {
+      if (
+        inputType !== RANGE_INPUT.MAX &&
```
nenonja, line 94: What if `inputType === RANGE_INPUT.ALL` ?

src/common/components/rangeDropdown/RangeDropdown.tsx
nenonja: another try:
```
const normalize = (val: string) => Number(val) || null;
const getValidatedMinValue = (val:string) => {
const values = [val, maxInputValue , maxInputEndValue].map(normalize);
return Math.min(...values);
}

```

src/common/components/rangeDropdown/__tests__/RangeDropdown.test.tsx
```diff
@@ -184,3 +190,75 @@ test('should call onChange correctly when setting fixed values with checkbox', (
...
+    userEvent.tab();
+    await waitFor(() => {
+      expect(onChange).toHaveBeenCalledWith('5', '5');
+    });
+
+    rerender({ minInputValue: '5', maxInputValue: '10', onChange });
+
+    userEvent.tab();
+    await waitFor(() => {
+      expect(onChange).toHaveBeenCalledWith('5', '5');
```
nenonja, line 261: SHouldnt this be `expect(onChange).toHaveBeenCalledWith('5', '10');`?

src/domain/courseSearch/Search.tsx
melniiv: cuz hds textinput accepts string...

src/common/components/rangeDropdown/RangeDropdown.tsx
melniiv: actually Santeri pointed out that there are a lot of code duplicates if its separated into 2 methods. So i combined it to one.

src/common/components/rangeDropdown/RangeDropdown.tsx
melniiv: No, its actually for RANGE_INPUT.MIN method.. so if min value > max value we fix the min value. Because its focused out, we can change only this field, and not control another field. ALL will just execute both min and max checks on component load.

src/common/components/rangeDropdown/__tests__/RangeDropdown.test.tsx
melniiv: yes, its not valid one.. fixed

src/common/components/rangeDropdown/RangeDropdown.tsx
melniiv: and i agree that it might be not the best implementation, i was refactoring it couple of times, usually finding new use cases (in which order use fills fields, on load values, on search, from keyboard or typing or mouse, etc).

src/common/components/rangeDropdown/RangeDropdown.tsx
nenonja: What I meant is that when `inputType === RANGE_INPUT.ALL` it also goes to this if-clause.

src/common/components/rangeDropdown/RangeDropdown.tsx
Sartonon: if @nenonja suggestion works as expected it seems good and easily understandable, of course that can be combine into 1 function also

melniiv: As i understand the logic of his approach and what i wanted to achieve is different. His function takes 3 strings, normalizes to int values and calculates Min or Max of the set of 3. But that's not what I wanted. we have a range of allowed numbers minAllowed current maxAllowed. So for Min and Max field we want to first put the current value within the range, if its outside the range currently. So if current selected field value (works for both min and max fields) less than min => current = min. If more than Max => current = max. And then we have the third condition, and logic will differ for 2 fields. Min value should be always less than Max. We check it on blur, so depending on the selected field we will correct the values differently. So if Min is selected and its more than current Max. then we correcting Min to be the same as current Max. In opposite, when Max field is selected, and current Min value > current Max, then we will fix current Max to be equal to current Min. So the order of the filling values actually matters. As we don't want to control/change other fields value on blur, only the selected field. Hope it makes sense.

src/common/components/rangeDropdown/RangeDropdown.tsx
melniiv: can you paste the code please

src/common/components/rangeDropdown/RangeDropdown.tsx
Sartonon: Could we just get rid of minInputStartValue prop, same for maxInputStartValue and just use initial value given to the component. Would that simplify this little bit?

Sartonon: To me @nenonja s suggestion works correctly for the most part but onyl doesn't work when it is called like this:

`handleInputBlur(RANGE_INPUT.ALL, '')`

Sartonon:
```
    const normalize = (val: string) => Number(val) || null;
    const getValidatedMinValue = (val: string) => {
      const values = [val, maxInputValue, maxInputEndValue]
        .map(normalize)
        .filter((n) => n) as number[];
      return Math.min(...values).toString();
    };
    const getValidatedMaxValue = (val: string) => {
      const values = [minInputStartValue, minInputValue, val]
        .map(normalize)
        .filter((n) => n) as number[];
      return Math.max(...values).toString();
    };
```

I just did little bit modifications to get it work

Sartonon: I bet it could be simplified further but I don't have time now

nenonja: perhaps it needs also check that the value is not out of range:
`Math.max(0,Math.min(...values))`
`Math.min(Math.max(...values), maxInputEndValue)`

PR #204, opened 2021-02-04
Th-992: hide collections that have live value false

nenonja:
## Description :sparkles:
collections have live attribute that should be always true when collection is visible on collections page, frontpage, similar collections page or included to sitemap.xml. Single collection page is excluded, becuse it can parameterized by 
?draft=(true|false)
See conversations: 
- https://helsinkicity.slack.com/archives/C013682SYSD/p1612424363037400
- https://helsinkicity.slack.com/archives/C9N9KNYM8/p1612271522002600
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-992
## Testing :alembic:
create a collection to wagtail. Publish it, check that it is visible in all above pages and sitemap. Unpublish the collection. Now it should be invisible. Previously it was not.


Sartonon (commented):
Looks good. Should we create new folder for test specific utility functions / files? I think it would be nice.

src/util/collections.common.tests.ts
```diff
@@ -0,0 +1,150 @@
...
+  });
+};
+
+const renderComponentAndExpectOddCollectionsToBePresent = async (
+  component: React.ReactElement,
+  collections: CollectionListResponse,
+  collectionNames: string[],
+  mocks: MockedResponse[] = [],
+  variables?: CollectionListQueryVariables
+) => {
```
Sartonon, line 46: I would prefer these functions would receive one object as parameter instead of 4-5 parameters, It would easier in the future to pass parameters because the ordering doesn't have to be anything specific.

With destructuring passing object is handy

src/domain/collection/CollectionUtils.ts
```diff
@@ -56,6 +56,23 @@ export const isCollectionExpired = (
...
+  return Boolean(collection.live);
+};
+
+export const isCollectionVisible = (
+  collection: CollectionFieldsFragment,
+  locale: Language
+): boolean => {
+  return (
+    isLanguageSupported(collection, locale) &&
+    !isCollectionExpired(collection) &&
```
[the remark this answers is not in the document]
nenonja: Code can always be cleaner, done.

src/util/collections.common.tests.ts
nenonja: Yes it makes it also more readable at least when browsing github. Done.

Sartonon (approved):
Looks good

PR #210, opened 2021-02-15
TH-1007: Use query parameter to memorize return path when clicking back button from event page

nenonja:
Let's use query parameter "returnPath" to memorize previous page when routing from eventcard to single event page.
It works on every page which contains event card. If queryparam is missing, return to event search page as a fallback.
There is a special case when follows similar events multiple times. In this case queryParams remembers the history of return paths correctly by having multiple parameters. Note! To test this special case, you need to enable "SimarEvents" feature flag becuase currently they are hidden.
On Arial-label of the button, the old text is obsolete "Back to event search", was changed to "Go back to previous page". The text could be dynamic based on the returnPath but that is too much investment of work against the benefits :)
Sharing links seem to work out of the box in a way that it strips the query parameter which is good.

## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-1007


src/domain/event/eventQueryString.util.ts
```diff
@@ -0,0 +1,30 @@
...
+export type ReturnParams = {
+  returnPath: string;
+  remainingQueryString?: string;
+};
+
+/**
+ * Extracts latest return path from queryString. For example on:
+ * http://localhost:3000/fi/event/kulke:53397?returnPath=%2Fevents&returnPath=%2Fevent%2Fhelsinki%3Aaf3pnza3zi
+ * latest return path is in the last returnPath param on queryString : %2Fevent%2Fhelsinki%3Aaf3pnza3zi
+ * @param queryString
```
Sartonon, line 12: does this line add anything useful, could it be deleted?

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
```diff
@@ -60,17 +68,24 @@ afterAll(() => {
   clear();
 });
 
-test('should render similar event cards', async () => {
+const waitForComponentToBeLoaded = async (render: () => CustomRenderResult) => {
```
Sartonon, line 71: can't quite see the benefit passing the render as a callback here

Sartonon:
```
render(<SimilarEvents event={event} />, { mocks })
await waitForComponentToBeLoaded();
```

that seems easier to me

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
```diff
@@ -83,3 +98,30 @@ test('should render similar event cards', async () => {
...
+it('has return path on similar event link', async () => {
+  const path = ROUTES.EVENT;
+  const route = path.replace(':id', event.id);
+  const { history } = await waitForComponentToBeLoaded(() =>
+    renderWithRoute(<SimilarEvents event={event} />, {
+      mocks,
+      path,
+      routes: [route],
+    })
+  );
```
Sartonon, line 111: not sure why `renderWithRoute` is used in this test scenario. I think this is meant to be used with components that are actual route components (top level page component)

this works also here
```
 const { history } = await waitForComponentToBeLoaded(() =>
    render(<SimilarEvents event={event} />, {
      mocks,
      routes: [route],
    })
  );
```

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
```diff
@@ -83,3 +98,30 @@ test('should render similar event cards', async () => {
...
+          '{{name}}',
+          similarEvent.name.fi
+        ),
+      })
+    );
+    expect(history.location).toMatchObject({
+      pathname: `/fi${ROUTES.EVENT.replace(':id', similarEvent.id)}`,
+      search: `?returnPath=${encodeURIComponent(route)}`,
+    });
+    history.goBack();
```
Sartonon, line 125: Are you aware that this works without further waiting because data comes instantly from cache (not async)?

for example if I change EventListQuery to:

```
  const { data: eventsData, loading } = useEventListQuery({
    ssr: false,
    variables: eventFilters,
    fetchPolicy: 'network-only',
  });
```

this breaks

src/util/queryString.ts
```diff
@@ -0,0 +1,47 @@
...
+    case 'returnPath':
+      return stripLanguageFromPath(value);
+    default:
+      return assertUnreachable(param, 'Unknown query parameter');
+  }
+};
+
+export const addEntriesToQueryString = (
+  queryString: string,
+  ...entries: Array<QueryEntry>
```
Sartonon, line 33: The way these entries are passed bothers me little bit:

`addEntriesToQueryString('string', {param, value}, {param, value}, {param, value}, {param, value}, ...)`

Would it be better to just have single object with key-value pairs?

```
addEntriesToQueryString(
'string', 
{ places: 'param', returnPath: 'param }  
 ...
 )
```

Sartonon (commented):
The way this functionality is implemented looks very good to me. Some comments but they aren't that much related to the implementation

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
Sartonon: Of course a comment could be added to renderWithRoute utility to clarify this

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
Sartonon: I would try to make this test more resilient by adding wating functionality to every iteration of this for loop

src/domain/event/eventQueryString.util.ts
nikomakela: > does this line add anything useful, could it be deleted?

An URL example is a good addition. It helped me to understand whats the input to this function.

Sartonon: I was only referring to line 12:

 * @param queryString

Sartonon: Example is good 👍

nenonja: Removed. IntelliJ added it automatically.

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
nenonja: At some point it was benefical but now unnedessary complex as nothing is done with rendering result inside the method. I can fix it.

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
nenonja: That's strange, it did not work previously with `render` but now it does. I thought that the router is needed for the history operation. But I did originally copy-cat this from https://github.com/City-of-Helsinki/events-helsinki-ui/blob/master/src/domain/event/eventCard/__tests__/EventCard.test.tsx#L88-L125

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
nenonja: Good point, not sure did I do it now correctly, I added the loadingspinner wait inside the loop and the test wanted it also to wrapped with `act` but now it passed with that network-only policy

src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx
Sartonon: I need to check if EventCard.test.tsx needs renderWithRoute

src/util/queryString.ts
nenonja: The idea was for the method to be common enough to accept any type of atributes, also queryParams can also have multiple parameters of the same type, like:
```  
addEntriesToQueryString(
    'string',
    { param: 'returnPath', value: 'events' },
    { param: 'returnPath', value: '/event/helsinki:af3pnza3zi' }
  );
  ```

File: src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx (hunk 3 above)
Comment (Sartonon) [reply src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx:111]: seems like EventCard test needs renderWithRoute to get this hook return params:

`  const params = useParams<{ place?: string }>();`

File: src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx (hunk 4 above)
Comment (Sartonon) [reply src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx:125]: You haven't pushed any changes yet?

File: src/util/queryString.ts (hunk 5 above)
Comment (Sartonon) [reply src/util/queryString.ts:33]: With multiple parameters of the same type I think array would do. But maybe my opinion comes from how some other query params builders work in other libraries

Comment (nenonja) [reply src/util/queryString.ts]: Ah I thought I did not send the previous message when I started to investigate further. Now there is resolution with Object type and value can be either string or array of strings.

File: src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx (hunk 4 above)
Comment (nenonja) [reply src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx]: now pushed, the fixing the last comment took some time!

Comment (Sartonon) [reply src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx]: I was just making sure because that one time you forgot to push :D

File: src/domain/event/similarEvents/__tests__/SimilarEvents.test.tsx [hunk 6]
```diff
@@ -83,3 +88,32 @@ test('should render similar event cards', async () => {
...
+        })
+      );
+      expect(history.location).toMatchObject({
+        pathname: `/fi${ROUTES.EVENT.replace(':id', similarEvent.id)}`,
+        search: `?returnPath=${encodeURIComponent(route)}`,
+      });
+      history.goBack();
+    });
+    await waitForComponentToBeLoaded();
+  }
```
Sartonon, line 118: usually when using act, you are trying to hide some underlying problem. I tested that this version works without act:

```
  await waitForComponentToBeLoaded();
  for (const similarEvent of expectedSimilarEvents.data) {
      userEvent.click(
        screen.queryByRole('button', {
          name: translations.event.eventCard.ariaLabelLink.replace(
            '{{name}}',
            similarEvent.name.fi
          ),
        })
      );
      expect(history.location).toMatchObject({
        pathname: `/fi${ROUTES.EVENT.replace(':id', similarEvent.id)}`,
        search: `?returnPath=${encodeURIComponent(route)}`,
      });
      history.goBack();
      await waitForComponentToBeLoaded();
  }
```

Sartonon: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

nenonja: Last extra await in the end of the loop was left there without reason, it seem to work without it. Fixed now. Good to know that suggesting using `act` does mean you should use it :)

Sartonon: Yes it is little bit confusing but we are actually using act but it is react testing library that is doing it behind the scenes, so most of the time we don't need to write it ourselves

Sartonon: "You can read more about React's act utility here, but again, you shouldn't have to use it very often. It's built-into React Testing Library. There are very few times you should have to use it directly if you're using React Testing Library's async utilities."

Sartonon (approved):

PR #211, opened 2021-02-15
TH-999: Separate publisher and provider info

melniiv:


nikomakela:
This pull request is so much out of date that I started a new one #273. The commit from this one is taken as a base for it.

PR #213, opened 2021-02-16
TH-1006: Show super event date in event hero if exists

melniiv:


src/domain/course/CoursePageContainer.tsx
```diff
@@ -55,7 +81,11 @@ const CoursePageContainer: React.FC = () => {
                 <EventClosedHero />
               ) : (
                 <>
-                  <EventHero event={course} eventType="course" />
+                  <EventHero
+                    event={course}
+                    superEvent={superEvent?.courseDetails}
```
nenonja, line 86: I was thinking that what if we would pass superEventId to the `EventHero` and doing query inside the EventHero? So we would avoid duplicate code in `CoursePageContainer` and `EventPageContainer`. Another option to avoid duplicate code is to make custom hook for loading superEvent and use it in both files.

src/domain/event/eventHero/EventHero.tsx
```diff
@@ -79,6 +80,9 @@ const EventHero: React.FC<Props> = ({ event, eventType }) => {
     }
   }, [imageUrl]);
 
+  const superEventStartTime = superEvent ? superEvent.startTime : startTime;
```
nenonja, line 83: `superEventStartTime` might be a misleading name as it is not necessary super event start time. Perhaps rename `startTime` to `eventStartTime` and use:
 `const startTime = superEvent ? superEvent.startTime : eventStartTime;`

src/domain/event/eventHero/EventHero.tsx
```diff
@@ -111,10 +115,11 @@ const EventHero: React.FC<Props> = ({ event, eventType }) => {
                 <div className={styles.description}>{shortDescription}</div>
               )}
               <Visible above="sm" className={styles.date}>
-                {!!startTime &&
+                {superEvent !== undefined &&
```
nenonja, line 118: I didnt test but quickly it looks like that not any date range is shown when superevent is not defined? Shouldnt it default the to the event time then?

nenonja: Also it seems that no test goes red because of this. There should be a two tests on EventHero with and without the superEvent and both checks that the value of dateRange is correct

nenonja (changes requested):
I left some comments, also I think that shouldnt this feature be branched from `master` instead of `events-helsinki-2`? I think this is wanted to the production before 2.0.
If it's so, basically this feature could be done in two pr:s, first on master and then another in events-helsinki-2 taht adds the super event date to the coursePage.

nenonja:
It seems that Tomi Aalto's input about the resolution is still missing from the ticket. Maybe we should wait for it before merging...

src/domain/event/eventHero/EventHero.tsx
nenonja: Yup it seems to be a regression bug introduced, so tests should be added.
Screenshot 2021-02-16 at 2 18 46 PM

src/domain/course/CoursePageContainer.tsx
```diff
@@ -39,8 +44,29 @@ const CoursePageContainer: React.FC = () => {
...
+      include: ['in_language', 'keywords', 'location', 'audience'],
+    },
+  });
+
   const course = courseData?.courseDetails;
 
+  useEffect(() => {
+    if (course) {
+      setSuperEventId(
+        last(course.superEvent?.internalId?.split('/').filter((e) => e)) || ''
```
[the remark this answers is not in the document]
melniiv: good point

src/domain/course/CoursePageContainer.tsx
melniiv: well, its extra query, but we might need the superevent data in other components. @Sartonon, what do you think?

src/domain/event/eventHero/EventHero.tsx
melniiv: ok

src/domain/event/eventHero/EventHero.tsx
melniiv: i will check

src/domain/course/CoursePageContainer.tsx
melniiv: we have the same duplicate query of the event itself basically

src/domain/event/eventHero/EventHero.tsx
melniiv: fixed

melniiv:
yes, with loading part we need Tomi, i ll request to test

src/domain/event/eventHero/EventHero.tsx
nenonja: you can rename it:
```
{
...
startTime: eventStartTime,
endTIme: eventEndTime,
...
} = useEventQuery(...);
```

melniiv: i know that i can:) lets not overcomplicate

melniiv: fixed

src/domain/course/CoursePageContainer.tsx
```diff
@@ -39,8 +46,39 @@ const CoursePageContainer: React.FC = () => {
...
+
+  useEffect(() => {
+    if (superEventId) {
+      getData();
+    }
+  }, [getData, superEventId]);
+
+  useEffect(() => {
+    setSuperEventData(superEvent?.courseDetails || null);
+  }, [superEvent]);
```
Sartonon, line 80: Do we need to split this into 3 useEffects? Could this be handled in one?

melniiv: i think 3 needed. we can call getData only when we have superEventId, so we have to store it in state, as we get it from the first query. superEventData can be undefined (if no superevent id), null (when we dont know if there is superevent id, or when its loading) and the actual superEvent object. To handle all those 3, and not make EventHero blinking with wrong data, i think we have to keep logic like that. Effects just allow to make a sync sequence. But if you have better implementation, please tell:)

src/domain/event/eventHero/__tests__/EventHero.test.tsx
```diff
@@ -192,3 +193,42 @@ test('Register button should be visible and clickable', () => {
...
+  });
+  render(
+    <EventHero
+      event={mockEvent}
+      superEvent={mockSuperEvent}
+      eventType="course"
+    />
+  );
+
+  const superDateStr = getDateRangeStr({
```
nenonja, line 226: I would use hard coded string here for date range, if there is a bug in `getDateRangeStr` it is duplicated here

nenonja:
> I left some comments, also I think that shouldnt this feature be branched from `master` instead of `events-helsinki-2`? I think this is wanted to the production before 2.0.
> If it's so, basically this feature could be done in two pr:s, first on master and then another in events-helsinki-2 after merging master that adds the super event date to the coursePage.

Mari said that this will be only in Courses and 2.0.

src/domain/event/eventHero/__tests__/EventHero.test.tsx
melniiv: then we have to rewrite all tests which use getDateRangeStr, OtherCourseTimes etc.. do we want to do that as a part of this pr as well?

src/domain/course/CoursePageContainer.tsx
nenonja: There is some discussion about it: https://stackoverflow.com/questions/54002792/should-i-use-one-or-many-useeffect-in-component
I think first two useEffects fall to category 2 and should be separated.
But can you merge last two useEffects like this?
```
  useEffect(() => {
    if (superEventId) {
      const superEventData = getData();
      setSuperEventData(superEventData);
    }
  }, [getData, superEventId]);
  ```

Comment (nenonja) [reply src/domain/course/CoursePageContainer.tsx:80]: Also I would still put them all to custom hook, it's more readable and more reusable (if needed elsewhere later)

File: src/domain/event/eventHero/__tests__/EventHero.test.tsx (hunk 6 above)
Comment (nenonja) [reply src/domain/event/eventHero/__tests__/EventHero.test.tsx:226]: Well I just noticed that the getDateRangeStr.ts has unit tests which covers possible bugs. So No need to fix this one if you dont feel like doing it. I just would find the test more readable when using hard coded values.

File: src/domain/course/CoursePageContainer.tsx (hunk 5 above)
Comment (Sartonon) [reply src/domain/course/CoursePageContainer.tsx:80]: If we really need to have separate state for superEventData. I wouldn't use lazy hook in the first place.

**Comment by nenonja** [comment]
There could be loading-spinner in place when still loading the event time to avoid moving components on page.

File: src/domain/event/eventHero/__tests__/EventHero.test.tsx (hunk 6 above)
Comment (melniiv) [reply src/domain/event/eventHero/__tests__/EventHero.test.tsx:226]: I am just thinking what if the implementation will change inside the getDateRangeStr, and thats the method we use inside the EventHero itself. So tests will fail with hardcoded values then

**Comment by melniiv** [comment]
yes, we are discussing now with Tomi, he said that spinner might be overkill, but some animation, he is preparing the example

File: src/domain/course/CoursePageContainer.tsx (hunk 5 above)
Comment (melniiv) [reply src/domain/course/CoursePageContainer.tsx]: what is the solution without lazy hook and why it is bad?

File: src/domain/course/CoursePageContainer.tsx [hunk 7]
```diff
@@ -41,6 +49,29 @@ const CoursePageContainer: React.FC = () => {
 
   const course = courseData?.courseDetails;
 
+  const superEventId = getEventIdFromUrl(course?.superEvent?.internalId ?? '');
```
nenonja, line 52: shouldnt this be `getEventIdFromUrl(course?.superEvent?.internalId) ?? '';`

package.json
```diff
@@ -61,6 +61,7 @@
     "react": "^16.13.1",
     "react-apollo": "^3.1.3",
     "react-apollo-hooks": "^0.5.0",
+    "react-content-loader": "^6.0.1",
```
nenonja, line 64: hds supports small spinner, I think we should use that:  https://hds.hel.fi/components/spinner what do you think? It's best to ask about it from Tomi.
If we start to use hds spinner, we can delete our own spinner component completely, just remember then to put `data-testid="loading-spinner"` on it or many tests will fail.

nenonja (commented):
Some more comments

nenonja:
I just fixed browser-tests and they rely more on loading spinner, so rebasing events-helsinki-2 would be help to make tests green and avoid conflicts with spinner.

src/domain/course/CoursePageContainer.tsx
melniiv: no, getEventIdFromUrl cannot accept undefined

package.json
melniiv: Tomi said that we need skeleton, spinner is not that good in that case. So i made a demo for him and he approved now.

nenonja:
react-content-loader has cpu issue, but perhaps not deal-breaker: https://github.com/danilowoz/react-content-loader/issues/125

nenonja (approved):
If Tomi approved the ui, this is fine for me too 👍

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
```diff
@@ -51,7 +52,15 @@ const OtherEventTimes: React.FC<Props> = ({
...
-  if (!superEventId || events.length === 0) return null;
+  if (!superEventId) {
+    return null;
+  } else if (events.length === 0 && loading) {
+    return (
+      <div className={styles.skeletonWrapper}>
+        <SkeletonLoader />
+      </div>
+    );
+  }
```
Sartonon, line 63: After this change we are showing "show other event times" button when events.length === 0

Sartonon: Should be `if (!superEventId || events.length === 0) {` ?

melniiv: what is the difference from previous implementation except {}? the whole point of the fix to add loader, Tomi suggested. if we leave previous, then loader will never appear

melniiv: if there is no super id, then null always. if there is superid, then probably there will be other times loaded.. no?

Sartonon: It is possible that there are no other events even when superEventId is available. Like in the screenshot

Sartonon: Previously it wouldn't show that section at all

src/domain/course/__tests__/CoursePageContainer.test.tsx
```diff
@@ -243,3 +281,29 @@ it('should link to courses search when clicking tags', async () => {
...
+    start: superEventResponse.data.courseDetails.startTime,
+    end: superEventResponse.data.courseDetails.endTime,
+    locale: 'fi',
+    includeTime: true,
+    timeAbbreviation: translations.commons.timeAbbreviation,
+  });
+
+  expect(
+    screen.getAllByText((_content, el) => el.textContent === superDateStr)
+  ).toHaveLength(1);
```
Sartonon, line 308: This would work also:
``` 
 expect(
    screen.getByText((_content, el) => el.textContent === superDateStr)
  ).toBeInTheDocument();
```

Also adding data-testid and querying with that might be helpful.

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
```diff
@@ -51,7 +52,15 @@ const OtherEventTimes: React.FC<Props> = ({
     history.push(eventUrl);
   };
 
-  if (!superEventId || events.length === 0) return null;
+  if (!superEventId || (events.length === 0 && !loading)) {
```
Sartonon, line 55: Hmm to me this seems quite complex if statement for pretty simple thing, I think there is a way to make this much simpler and easier to understand

Sartonon: hmm why is superEventId provided in the first place as a prop? I have no idea why I have done that

Sartonon: I think that the superEventId prop can be omitted and then the code woul be something like this:

```
  if (loading) {
    return (
      <div className={styles.skeletonWrapper}>
        <SkeletonLoader />
      </div>
    );
  }

  if (events.length === 0) {
    return null;
  }

```

melniiv: i think its just logical: If superEventId is not provided -> do nothing, and then only some extra checks.

src/domain/course/__tests__/CoursePageContainer.test.tsx
melniiv: toBeInTheDocument() does not work, because returned type is array with one element.

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
melniiv: i can fix it if its more readable though. and yes, in that case superEventId can be removed

src/domain/course/__tests__/CoursePageContainer.test.tsx
melniiv: fixed

src/domain/event/eventInfo/otherEventTimes/OtherEventTimes.tsx
melniiv: fixed

Sartonon (approved):
Good :)

PR #215, opened 2021-02-19
TH-1016 | Add tests for pagination in CuratedEventList

Sartonon:
## Description :sparkles:

- Add tests for pagination in CuratedEventList

## Issues :bug:

**[TH-1016](https://helsinkisolutionoffice.atlassian.net/browse/TH-1016):**

### DIFF
--- a/src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
+++ b/src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
@@ -36,6 +36,7 @@ const getMocks = (events: EventFieldsFragment[], ids = eventIds) => [
   },
 ];
 
+// Creates array of mocks to match pagination queries
 const getMocksForPagination = (eventsCount = 35) => {
   const eventIds = range(eventsCount).map((i) => (i + 1).toString());
   const eventNames = eventIds.map((id) => `Event ${id}`);
@@ -162,6 +163,7 @@ test('event list pagination works', async () => {
     userEvent.click(screen.getByRole('button', { name, hidden: true }));
   };
 
+  // use loop to fetch all the events by clicking show more button (pagination)
   for (
     let eventsFetchedCount = PAGE_SIZE;
     eventsFetchedCount < eventsCount;
--- a/src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
+++ b/src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
@@ -1,4 +1,6 @@
 import { advanceTo, clear } from 'jest-date-mock';
+import chunk from 'lodash/chunk';
+import range from 'lodash/range';
 import React from 'react';
 
 import translations from '../../../../common/translation/i18n/fi.json';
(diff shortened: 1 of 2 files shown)


src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
```diff
@@ -21,19 +23,62 @@ const collection = fakeCollection({
...
   },
 ];
 
+// Creates array of mocks to match pagination queries
+const getMocksForPagination = (eventsCount = 35) => {
+  const eventIds = range(eventsCount).map((i) => (i + 1).toString());
+  const eventNames = eventIds.map((id) => `Event ${id}`);
+  const curatedEvents = eventIds.map(
+    (id) => `http://localhost:3000/fi/event/${id}`
+  );
```
melniiv, line 45: looks like it can be only one map instead of three, but might be not that readable then.

Sartonon: eventIds for example are needed later in the code. Maybe the could be array that that has objects that contains eventId, eventName and curatedEvents

Sartonon: but then again I don't know if it makes it less readable

src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
```diff
@@ -21,19 +23,62 @@ const collection = fakeCollection({
...
+  });
+
+  const collection = fakeCollection({
+    curatedEvents,
+  }) as CollectionFieldsFragment;
+
+  return { mocks, collection, eventNames };
+};
+
+const waitForRequestToComplete = async () => {
```
nenonja, line 76: at some point this could be moved to some util method as it is used like 100s of different tests :)

src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
```diff
@@ -104,3 +149,40 @@ test('should show expired events', async () => {
...
+      )
+    );
+    await waitForRequestToComplete();
+  }
+
+  expect(
+    screen.queryByRole('button', { name: /näytä lisää tapahtumia/ })
+  ).not.toBeInTheDocument();
+
+  for (const eventName of eventNames) {
```
nenonja, line 185: It seems that we keep on clicking the showMorebutton until it vanishes and then check that all events are present. What about checking the event names that belongs to particular chunk, that they are present (inside the loop)?
Perhaps it does not matter except if events would somehow would come in unexpected order, then this would not catch it

nenonja (approved):
Some small things I commented but the tests seem to do the job, so approved.

src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
Sartonon: yeah at least for the loading spinner

src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
Sartonon: refactored the function little bit by using flow

src/domain/collection/curatedEventList/__tests__/CuratedEventList.test.tsx
Sartonon: Added tests to the pagination loop

PR #216, opened 2021-02-22
TH-990 | Single hobby page fixes

Sartonon:
## Description :sparkles:

- Fix price formatting for hobbies

### DIFF
--- a/src/domain/event/EventUtils.ts
+++ b/src/domain/event/EventUtils.ts
@@ -78,7 +78,7 @@ export const formatPrice = (price?: string): string => {
     return '';
   }
 
-  const priceRegex = /^\d+([/\-.,]{1})?\d+?$/;
+  const priceRegex = /^\d+([/\-.,]{1}\d+)?$/;
   return price.match(priceRegex) ? `${price} €` : price;
 };
 
--- a/src/domain/event/EventUtils.ts
+++ b/src/domain/event/EventUtils.ts
@@ -78,7 +78,7 @@ export const formatPrice = (price?: string): string => {
     return '';
   }
 
-  const priceRegex = /^\d+([/\-.,]{1}\d+)?$/;
+  const priceRegex = /^\d+([/\-.,]\d+)?$/;
   return price.match(priceRegex) ? `${price} €` : price;
 };
 
--- a/src/domain/event/EventUtils.ts
+++ b/src/domain/event/EventUtils.ts
@@ -64,6 +64,24 @@ export const getEventIdFromUrl = (url: string): string | null => {
   return eventId?.[1].replace('/', '') || null;
 };
 
+/*
+ * Format string to price format (add €) if it is a number and is missing currency
+ * For example:
+ * 'random text' -> 'random text'
+ * '2' -> '2 €'
+ * '2.5' -> '2.5 €'
+ * '30/50' -> '30/50 €'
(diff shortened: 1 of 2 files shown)


src/domain/event/EventUtils.ts
```diff
@@ -64,6 +64,24 @@ export const getEventIdFromUrl = (url: string): string | null => {
...
+ * '30/50' -> '30/50 €'
+ * '30-50' -> '30-50 €'
+ */
+export const formatPrice = (price?: string): string => {
+  if (!price) {
+    return '';
+  }
+
+  const priceRegex = /^\d+([/\-.,]{1})?\d+?$/;
+  return price.match(priceRegex) ? `${price} €` : price;
```
Sartonon, line 82: Open to hear other ideas how to achieve this with regex. Now many price infos are coming without euro sign.

nikomakela: In my opinion that looks good enough, even that regex is always hard to understand at first sight.

You could check with something like endsWith, but it's quite cryptic also..:
```
a += !a.endsWith("€") ? " €" : ""
b = !b.endsWith("€") || b + " €"
if (!c.endsWith("€")) c += " €"
```

Sartonon: We need to also know if its number and allow / - , . in the middle :) Because regex is hard to understand I added comments to the function definition. But I could also add comment to the regex itself

Sartonon: and of course no need to add € if it is just a description. We don't know exactly what is coming from linkedevents :)

nenonja: I tried it here: https://regex101.com/
Tthe site complaisn regex being broken: `/ An unescaped delimiter must be escaped with a backslash (\)`
So perhaps `^\d+([\/\-.,]{1})?\d+?$` is correct?
Also if it has only one number it does not work: "2" -> "2". 
Perhaps more unit tests are needed ;)

nenonja (changes requested):
One fix needed

src/domain/event/EventUtils.ts
Sartonon: well actually I removed those because eslint complained and it was still working :D

Sartonon: Screenshot 2021-02-24 at 19 01 21
If you change the flavor to ecmascript it doesn't  complain

Sartonon: Anyway, fixed and more tests added

src/domain/event/EventUtils.ts
```diff
@@ -78,7 +78,7 @@ export const formatPrice = (price?: string): string => {
     return '';
   }
 
-  const priceRegex = /^\d+([/\-.,]{1})?\d+?$/;
+  const priceRegex = /^\d+([/\-.,]{1}\d+)?$/;
```
nenonja, line 81: https://regex101.com/ says: `{1} Quantifier — Matches exactly one time (meaningless quantifier)`
so this {1} is probably redunant and could be removed,

Sartonon: true

Sartonon: fixed

nenonja (approved):

PR #217, opened 2021-02-22
TH-1054, TH934: Update HDS to 0.23.0 to fix header SSR issue

nenonja:
## Description :sparkles:

- Update HDS to 0.23.0 / Fix Header component for SSR
- New mock funcs


src/common/components/ssr/__tests__/VisibleAfterComponentMounted.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+it('is visible after mounting', async () => {
+  const Component = (
+    <VisibleAfterComponentMounted>Hello World</VisibleAfterComponentMounted>
+  );
+  const elem = create(Component);
+  expect(elem.root.findByType('span').props.style).toEqual({
+    visibility: 'hidden',
+  });
+  elem.update(Component);
```
Sartonon, line 17: came across this: https://reactjs.org/docs/test-renderer.html#testrendereract

Any idea why it is recommened to wrap these with act? Their first example doesn't have it

Sartonon: and thinking if `elem` coul use better naming? React docs use testRenderer and testInstance === testRendered.root

src/domain/banner/bannerHero/BannerHero.tsx
```diff
@@ -100,7 +101,8 @@ const BannerHero: React.FC<Props> = ({ banner, location }) => {
   const testIds = getTestIds(location);
 
   return (
-    <div
+    <VisibleAfterComponentMounted
```
Sartonon, line 104: Are we going to need these to get around ssr problems in HDS or?

nenonja: In a way yes, without this the banner image and text will jump around before their final position (after client side render). The idea is to hide the component until it is mounted. With "Visiblity: hidden" the component still would reserve the same space even when not visible.

nenonja:
Seems to work perfectly 👍 I will merge this now.

PR #218, opened 2021-02-23
Fix/th 975 remove dup styles

nikomakela:
## Description :sparkles:
Removed some duplicate style imports.

## Issues :bug:
TH-975 Investigated why the normalize.css was added 3 times inside the head-section and it looked like there are lot of deprecated stuff which now come from HDS or other plugins.

Removing these styles seems to have no effect on active styles. However, I don't know the build system or this product yet, so this needs to be carefully reviewed.

More comments in task: https://helsinkisolutionoffice.atlassian.net/browse/TH-975


Sartonon:
Now that I'm looking more into this seems like in prod build and with server side rendering we are getting all styles twice :) Even normalize appears twice with these changes :D

nikomakela:
Regarding to slack messages, I reverted the changes I made to webpack configuration.

Sartonon:
This already improves the situation, approved :)

Sartonon (approved):

PR #223, opened 2021-03-01
TH-917 | Hobby page link in footer

Sartonon:
## Description :sparkles:

- Hobby page link in footer
- Improve footer tests


src/domain/app/footer/__tests__/Footer.test.tsx
```diff
@@ -1,10 +1,37 @@
...
+import { ROUTES } from '../../routes/constants';
 import Footer from '../Footer';
 
 test('matches snapshot', () => {
   const { container } = render(<Footer />);
 
   expect(container.firstChild).toMatchSnapshot();
 });
+
+test.only('clicking links should route to right place', () => {
```
nikomakela, line 13: Is the "only" there in a purpose?

nikomakela (changes requested):
test.only needs to be addressed. Even if it is wanted to run only 1 test, it's better to skip rest.

src/domain/app/footer/__tests__/Footer.test.tsx
Sartonon: I use it only to get faster feedback when testing/debugging one test

PR #224, opened 2021-03-02
TH-1022 | Add scroll restoration to collection page

Sartonon:
## Description :sparkles:

- Add scroll restoration to collection page

I spent quite a bit of time investigating how scroll positions could be handled and if we could have some reusable logic here but ended up with simple solution for now... If someone has good suggestions how to could be handled lets hear them :)

### DIFF
--- a/src/domain/collection/CollectionPageContainer.tsx
+++ b/src/domain/collection/CollectionPageContainer.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { useLocation, useParams } from 'react-router';
+import { useHistory, useLocation, useParams } from 'react-router';
 import { Link } from 'react-router-dom';
 
 import ErrorHero from '../../common/components/error/ErrorHero';
@@ -14,6 +14,7 @@ import useLocale from '../../hooks/useLocale';
 import MainContent from '../app/layout/MainContent';
 import PageWrapper from '../app/layout/PageWrapper';
 import { ROUTES } from '../app/routes/constants';
+import { getLargeEventCardId } from '../event/EventUtils';
 import CollectionHero from './collectionHero/CollectionHero';
 import styles from './collectionPage.module.scss';
 import CollectionPageMeta from './collectionPageMeta/CollectionPageMeta';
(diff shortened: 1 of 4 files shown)


Sartonon:
I spent quite a bit of time investigating how scroll positions could be handled and if we could have some reusable logic here but ended up with simple solution for now... If someone has good suggestions how to could be handled lets hear them :)

src/domain/collection/CollectionPageContainer.tsx
```diff
@@ -82,6 +85,24 @@ const CollectionPageContainer: React.FC = () => {
...
+
+  // Scroll back to correct event card (event tat user is coming back from)
+  React.useEffect(() => {
+    if (location.state?.eventId) {
+      const cardId = getLargeEventCardId(location.state.eventId);
+      scrollToEventCard(cardId);
+      // Clear eventId value to keep scroll position correctly
+      const state = { ...location.state };
+      delete state.eventId;
+      history.replace({ ...location, state });
```
Sartonon, line 102: Not sure in which situation this is really needed, it is same as in EventSearchPageContainer.tsx

nikomakela: Found this with a quick search: https://stackoverflow.com/a/47513369/784642

nikomakela (approved):
Looks good to me.

nikomakela:
Actually, I tested it from PR server and it seems to not work on my Chrome Version 88.0.4324.192 (Official Build) (x86_64). 

I clicked "Kolmen karhun..." 
image

I get this 
image

After browser's back button, I'm on top of the page!

nikomakela:
Okay, by clicking left-arrow -button it seems to work. 
image

src/domain/collection/CollectionPageContainer.tsx
Sartonon: I was just thinking in what situation this could cause problems if left out...

Sartonon: Without this if I would navigate back and forward I would still get restored scroll, not sure if that would be problematic

PR #225, opened 2021-03-02
TH-935 | Fix double 302 redirection

Sartonon:
## Description :sparkles:

- redirect "/" straight to "/:locale/home"

## Screenshots :camera_flash:

Screenshot 2021-03-02 at 19 57 24

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -40,11 +40,7 @@ export enum SUPPORT_LANGUAGES {
   EN = 'en',
 }
 
-export const supportedLanguages = [
-  SUPPORT_LANGUAGES.EN,
-  SUPPORT_LANGUAGES.FI,
-  SUPPORT_LANGUAGES.SV,
-];
+export const supportedLanguages = Object.values(SUPPORT_LANGUAGES);
 
 export enum DATE_PICKER_INPUT {
   START = 'start',
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -40,6 +40,12 @@ export enum SUPPORT_LANGUAGES {
   EN = 'en',
 }
 
+export const supportedLanguages = [
+  SUPPORT_LANGUAGES.EN,
+  SUPPORT_LANGUAGES.FI,
+  SUPPORT_LANGUAGES.SV,
+];
+
 export enum DATE_PICKER_INPUT {
   START = 'start',
   END = 'end',
--- a/src/server/__tests__/utils.test.tsx
+++ b/src/server/__tests__/utils.test.tsx
@@ -2,7 +2,7 @@
 import { handleReactRouterRedirection } from '../utils';
 
 describe('handleReactRouterRedirection', () => {
-  it('redirects to /fi/home 308', () => {
+  it("redirects '/fi' to '/fi/home' (308 Permanent Redirect)", () => {
     const redirect = jest.fn();
     const req: any = { path: '/fi' };
     const res: any = { redirect };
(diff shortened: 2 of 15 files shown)


nikomakela (commented):
Looks good to me, but I'm not very familiar with Router yet.

src/constants.ts
```diff
@@ -40,6 +40,12 @@ export enum SUPPORT_LANGUAGES {
   EN = 'en',
 }
 
+export const supportedLanguages = [
```
nenonja, line 43: `export const supportedLanguages = Object.values(SUPPORT_LANGUAGES);` is shorter and safer too if SUPPORT_LANGUAGES are updated?

nenonja: Also, there is dozens of `Object.values(SUPPORT_LANGUAGES)` in the codebase, they could be replaced with `supportedLanguages`

nenonja (commented):
I'm not guru with redirects at all but start reading about redirects:
https://www.drlinkcheck.com/blog/http-redirects-301-302-303-307-308#:~:text=Use%20a%20302%20redirect%20when,know%20what%20you%20are%20doing.
and it strongly suggest _not_ to use 308 code:
`Due to still limited support of 308, the recommendation is to always go with 301 redirects, unless you require POST requests to be redirected properly and are certain that all clients understand the 308 response code.`

Also I dont understand why `/fi -> /fi/home` is permanent redirect but `/` -> `/fi/home` is temporary redirect?

src/constants.ts
Sartonon: Yeah should use object.values

nenonja:
> `Due to still limited support of 308, the recommendation is to always go with 301 redirects, unless you require POST requests to be redirected properly and are certain that all clients understand the 308 response code.`

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308
It seems that only IE 11 with windows 7 or 8 or older does not work. Perhaps it's good enough for us.

nenonja:
Conflict must be resolved and hopefully browsert tests turn to green after rebasing, master has fixes to them.

Sartonon:
> I'm not guru with redirects at all but start reading about redirects:
> 
> https://www.drlinkcheck.com/blog/http-redirects-301-302-303-307-308#:~:text=Use%20a%20302%20redirect%20when,know%20what%20you%20are%20doing.
> 
> and it strongly suggest _not_ to use 308 code:
> 
> `Due to still limited support of 308, the recommendation is to always go with 301 redirects, unless you require POST requests to be redirected properly and are certain that all clients understand the 308 response code.`
> 
> 
> 
> Also I dont understand why `/fi -> /fi/home` is permanent redirect but `/` -> `/fi/home` is temporary redirect?
> 
> 

Me neither. That is what was told in SEO report.. maybe we need to ask them again

Sartonon:
Screenshot 2021-03-05 at 14 30 10

src/server/__tests__/utils.test.tsx
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { handleReactRouterRedirection } from '../utils';
+
+describe('handleReactRouterRedirection', () => {
+  it('redirects to /fi/home 308', () => {
```
nenonja, line 5: very small thing but description could be `"redirects '/fi' to '/fi/home' (308 Permanent Redirect)"` and 
`"redirects '/' to '/fi/home' (302 Temporary Redirect)"`

nenonja (approved):
Looks good! One small improvement suggestion, but not mandatory

src/server/__tests__/utils.test.tsx
Sartonon: fixed :)

PR #230, opened 2021-03-04
Small refactor

Sartonon:
## Description :sparkles:

- Regexp refactor
- Delete unused func

### DIFF
--- a/src/domain/event/EventUtils.ts
+++ b/src/domain/event/EventUtils.ts
@@ -1,5 +1,6 @@
 import { isPast, isThisWeek, isToday } from 'date-fns';
 import capitalize from 'lodash/capitalize';
+import sum from 'lodash/sum';
 
 import { EVENT_STATUS } from '../../constants';
 import {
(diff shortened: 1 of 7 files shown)


src/domain/event/EventUtils.ts
```diff
@@ -118,10 +116,7 @@ export const getEventPlaceholderImageUrl = (
   event: EventFieldsFragment
 ): string => {
   const numbers = event.id.match(/\d+/g);
```
nenonja, line 118: Super small thing but if number part of the id is needed somewhere else, this could be separate method like `const numbers = getNumberId(event);`

nenonja (approved):
Looks good, one very small comment that can be also ingnored 😉

src/domain/event/EventUtils.ts
Sartonon: Can't think of another place where this would be used right now so I'm not gonna abstract it at this point.

PR #232, opened 2021-03-05
Fix browser tests on events-helsinki-2 branch + removed doubled text from header

nenonja:
## Description :sparkles:
- fix browser tests
- removed doubled text from header (see screenshot)
## Screenshots :camera_flash:
Screenshot 2021-03-05 at 1 13 30 PM


nikomakela (approved):
Approved, but the logo -section could be redesigned, so it would not look like any of these:
- Helsinki Tapahtumat Tapahtumat Harrastukset
- Helsinki Tapahtumat Helsinki Tapahtumat Harrastukset

The links and the static logos and texts looks the same on top bar.

nenonja:
let's wait that the tests are green

PR #237, opened 2021-03-15
TH-1050: Date range min allowed date logic improvement

melniiv:


Sartonon (approved):
Seems to be working nicely, good job :)

PR #240, opened 2021-03-16
TH-1013; Footer redesign with dynamic categories

melniiv:
## Description :sparkles:
-Show events categories if current route is events
-Show courses categories if current route is courses
-Hide categories if route is generic (not courses or events)


src/domain/app/footer/FooterCategories.tsx
```diff
@@ -6,35 +7,58 @@ import CategoryFilter from '../../../common/components/category/CategoryFilter';
...
+  getCourseCategoryOptions,
   getEventCategoryOptions,
   getSearchQuery,
 } from '../../eventSearch/utils';
 import { ROUTES } from '../routes/constants';
 import styles from './footerCategories.module.scss';
 
-const FooterCategories: FunctionComponent = () => {
+interface FooterProps {
+  route: EventRouteProp;
```
Sartonon, line 25: `export type EventRouteProp = typeof ROUTES.EVENTS | typeof ROUTES.COURSES;`

would do the trick in EventRouteProp

src/domain/app/footer/Footer.tsx
```diff
@@ -16,13 +17,26 @@ const footerTheme = {
...
     document.querySelector<HTMLDivElement>(`#${resetFocusId}`)?.focus();
   };
 
+  const getCategoriesRoute = () => {
+    if (pathname.startsWith(`/${locale}${ROUTES.COURSES}`)) {
+      return ROUTES.COURSES;
+    } else if (pathname.startsWith(`/${locale}${ROUTES.EVENTS}`)) {
+      return ROUTES.EVENTS;
+    } else {
+      return '';
```
Sartonon, line 34: Maybe it would be more clearer if this just returned null or false when there is no category route

src/domain/app/footer/Footer.tsx
```diff
@@ -42,7 +56,7 @@ const FooterSection: FunctionComponent = () => {
           to={`/${locale}${ROUTES.COLLECTIONS}`}
         />
       </Footer.Navigation>
-      <FooterCategories />
+      {!!categoriesRoute && <FooterCategories route={categoriesRoute} />}
```
Sartonon, line 59: I think we can delete `!!` here

src/domain/app/footer/Footer.tsx
Sartonon: This could be implemented this way:

```
  const getCategoriesRoute = () => {
    return [ROUTES.COURSES, ROUTES.EVENTS].find((route) =>
      pathname.startsWith(`/${locale}${route}`)
    );
  };
```

Sartonon (commented):
Looking good butr added couple improvement ideas. Didn't test it yet as this is not updated yet to support api proxy update

src/domain/app/footer/Footer.tsx
melniiv: fixed

src/domain/app/footer/Footer.tsx
melniiv: fixed

src/domain/app/footer/FooterCategories.tsx
melniiv: fixed

Sartonon (approved):

PR #243, opened 2021-03-22
Moved BannerHero props to utils.

nikomakela:
TH-1069. TestCafe imported source code in wrong order. Moved BannerHero props to utils. Utils should not import jsx-component.


nenonja (approved):
Thsi resolves compilation problem in testcafe: one test cafe test imported bannerUtils which in turn imported BannerHero.tsx with all react stuff. Strange that also BannerHero.tsx imported bannerUtils which is cyclic dependency but it still worked... now luckily fixed! 👍

PR #244, opened 2021-03-22
TH-1043 | Support courses in collection page

Sartonon:
## Description :sparkles:

- Support course is collection page

## Screenshots :camera_flash:
Screenshot 2021-03-26 at 12 41 55


src/domain/collection/curatedEventList/CuratedEvents.tsx
```diff
@@ -0,0 +1,135 @@
...
+    <Container className={styles.container}>
+      {!!eventIds.length && (
+        <CollectionEventsList
+          testId={eventsListTestId}
+          eventIds={eventIds}
+          eventType="event"
+          title={t('collection.curatedEvents.eventsTitle')}
+        />
+      )}
+      {!!courseIds.length && (
```
nenonja, line 42: It seems that all the courses are always at the end of the collection, but in cms they can be in mixed order:
Screenshot 2021-03-26 at 1 54 58 PM
Is the collection designed to show them in two groups? This might be confusing to cms producer that the order has changed in the collection page...

src/domain/collection/curatedEventList/CuratedEvents.tsx
```diff
@@ -0,0 +1,135 @@
...
+    events,
+    pastEvents,
+    isFetchingMore,
+    loading,
+    onLoadMoreEvents,
+    hasMoreEventsToLoad,
+    eventCursorIndex,
+  } = usePaginatedEventsByIdsQuery(eventIds, eventType);
+
+  const visiblePastEvent = pastEvents.slice(
```
nenonja, line 73: `visiblePastEvents` might a bit better name

src/domain/collection/curatedEventList/CuratedEvents.tsx
```diff
@@ -0,0 +1,135 @@
...
+    setShowAllPastEvents(true);
+  };
+
+  return (
+    <div className={styles.curatedEventListContainer} data-testid={testId}>
+      <h3>{title}</h3>
+      <LoadingSpinner isLoading={loading}>
+        {(!!events.length || !!pastEvents.length) && (
+          <div className={styles.curatedEventList}>
+            <div>{!events.length && <OnlyExpiredEvents />}</div>
```
nenonja, line 88: Could this be shortened to:
```
 <div>{events.length ? <EventCards events={events} eventType={eventType}} /> : <OnlyExpiredEvents />}</div>

```

src/domain/collection/__tests__/CollectionPageContainer.test.tsx
```diff
@@ -48,7 +51,11 @@ const getMocks = (
...
   getEventsByIdsMock({
-    variables: { ids: [curatedEventId], include: ['location'] },
+    variables: {
+      ids: [curatedEventId],
+      include: ['location'],
+      source: LinkedEventsSource.Linkedevents,
+    },
     eventsByIds,
   }),
 ];
```
nenonja, line 61: For completeness there could be a test with collection with both events and collections and we would check that both lists are visible?

src/domain/collection/curatedEventList/CuratedEvents.tsx
nenonja: or even `visibleExpiredEvents` now we use words `expired` and `past` as synonyms and synonyms should be avoided

src/domain/collection/curatedEventList/EventCards.tsx
```diff
@@ -16,14 +18,15 @@ const EventCards: React.FC<Props> = ({
...
+  eventType,
 }) => {
   const { t } = useTranslation();
 
   return (
     <>
       <div className={styles.eventCards}>
         {events.map((event, id) => (
-          <LargeEventCard key={id} event={event} />
+          <LargeEventCard key={id} event={event} eventType={eventType} />
```
nenonja, line 29: strange that eventType was missing. I start to wonder now that do we event have to pass the type? Could we just detect the type from the event fields in some util function?

nenonja: Also wonder why is this added now, and was it a bug and some regression test missing?

src/domain/event/EventUtils.ts
```diff
@@ -57,9 +57,24 @@ export const isEventFree = (event: EventFieldsFragment): boolean => {
  * Get event id from url
  * For example  https://api.hel.fi/linkedcourses/v1/event/harrastushaku:13433?query -> harrastushaku:13433
  */
-export const getEventIdFromUrl = (url: string): string | null => {
-  const result = url.match(/\/event\/([^/?]*)/i);
-  return result?.[1] || null;
+export const getEventIdFromUrl = (
```
nenonja, line 60: A matter of taste but these could be in a RouterUtils.ts file or something like that? Most of the functions here are reducing results from EventFieldsFragment object. Also some tests could be added, unit tests are usually very easy to add.

src/domain/collection/curatedEventList/utils.ts
```diff
@@ -0,0 +1,104 @@
...
+
+import {
+  EventFieldsFragment,
+  LinkedEventsSource,
+  useEventsByIdsQuery,
+} from '../../../generated/graphql';
+import { isEventClosed } from '../../event/EventUtils';
+import { EventType } from '../../event/types';
+
+export const PAGE_SIZE = 10;
```
nenonja, line 13: exported but only used locally

src/domain/collection/curatedEventList/utils.ts
```diff
@@ -0,0 +1,104 @@
+import * as React from 'react';
```
nenonja, line 1: perhaps the name of the file could be more descriptive. If it's only exporting a hook, it could be even the name of the function: `usePaginatedEventsByIdsQuery.ts` or something shorter

nenonja (commented):
A lot of code and lot of comments too... I didnt test this yet.

src/domain/collection/curatedEventList/CuratedEvents.tsx
nenonja: The whole component is very long and hard to read, perhaps it could be refactored in smaller components?

src/domain/collection/curatedEventList/CuratedEvents.tsx
Sartonon: Idea is to show the in their own groups.

src/domain/collection/curatedEventList/CuratedEvents.tsx
Sartonon: yes should only use one naming for past/expired events

src/domain/collection/curatedEventList/EventCards.tsx
Sartonon: Nope, LargeEventCard had 'event' as default value. eventType is optional prop

src/domain/collection/curatedEventList/utils.ts
Sartonon: Seems to be duplicated, defined in two places

src/domain/event/EventUtils.ts
Sartonon: These are basically covered with integration/component tests. But maybe there are other benefits too to have tests for this small functions, faster feedback and easier to debug if big components tests are failing because of smaller functions

src/domain/collection/curatedEventList/utils.ts
Sartonon: Yes it is only exporting a hook atm but this is pretty much following the same pattern as other component folders, can't see any real benefit for a new file 🤔

src/domain/collection/__tests__/CollectionPageContainer.test.tsx
Sartonon: added course to "should show all events" test

PR #247, opened 2021-03-24
TH-957: Fix Autosuggestion Box Padding on mobile

nenonja:
## Description :sparkles:
TODO: Check usage of hds component

On front page screen the search input's autosuggestions box is completely hidden under android mobile keyboard. Let's fix this by making less padding between list items. At least few first items should be now visible. 
## Issues :bug:
https://helsinkisolutionoffice.atlassian.net/browse/TH-957
## Testing :alembic:
 This affects everywhere where input has autosuggestions, these all places should be tested. 
 If you have Android (or some other) phone, you perhaps can open the deployed site with it: http://events-helsinki-ui-th-957-fix-autosuggest-df.test.kuva.hel.ninja/
## Screenshots :camera_flash:
Screenshot 2021-03-24 at 9 30 31 AM

### DIFF
--- a/src/common/components/search/autosuggestMenu.module.scss
+++ b/src/common/components/search/autosuggestMenu.module.scss
@@ -1,3 +1,5 @@
+@import 'layout';
+
 .autosuggestMenu {
   --border-color-option: #f9f9f9;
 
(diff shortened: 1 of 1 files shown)


nenonja:
Not to be done this way. Let's use HDS component instead

PR #248, opened 2021-03-24
TH-929: Localized meta descriptions for landing pages.

melniiv:


src/common/translation/i18n/en.json
```diff
@@ -330,6 +330,20 @@
...
   },
+  "meta":{
+    "events": {
+      "description": "Tapahtumat.hel.fi’s event search. Concerts, gigs, theatre performances, exhibitions, culture, activities and other events in Helsinki."
+    },
+    "courses": {
+      "description": "Tapahtumat.hel.fi’s activity search. Activities, courses, clubs and camps for children and young people in Helsinki."
+    },
+    "collections": {
+      "description": "Check out our recommendations on the most interesting upcoming events."
```
nenonja, line 341: There supposed to be Exclamation mark `!` in the end of sentence 
Screenshot 2021-03-26 at 12 59 38 PM

src/common/translation/i18n/fi.json
```diff
@@ -330,6 +330,20 @@
...
+    "events": {
+      "description": "Tapahtumat.hel.fin tapahtumahaku. Konsertit, keikat, teatteriesitykset, näyttelyt, kulttuuri, tekeminen ja muut tapahtumat Helsingissä."
+    },
+    "courses": {
+      "description": "Tapahtumat.hel.fin harrastushaku. Lasten ja nuorten harrastukset, kurssit, kerhot ja leirit Helsingissä."
+    },
+    "collections": {
+      "description": "Tutustu suosituksiimme tämän hetken ajankohtaisimmista ja kiinnostavimmista tapahtumista."
+    },
+    "default": {
```
nenonja, line 343: Could find these `default` texts from the jira-ticket but perhaps these texts were determined somewhere else?

src/domain/landingPage/landingPageMeta/LandingPageMeta.tsx
```diff
@@ -1,39 +1,57 @@
...
     title: pageTitle,
   };
 
+  const getDescription = () => {
+    const route = [
+      ROUTES.COURSES,
+      ROUTES.EVENTS,
+      ROUTES.COLLECTIONS,
+    ].find((route) => pathname.startsWith(`/${locale}${route}`));
+    return t(`meta.${route?.substring(1) || 'default'}.description`);
```
nenonja, line 48: As far as I understand this is not what is wanted. This `LandingPageMeta` is part of the `LandingPage` and `LandingPagePreview`, so this should perhaps contain only default description. There is also `CollectionPageMeta` for collections page and `EventPageMeta` for courses and events pages. Also I found `StaticPageMeta` for accessibility and about page. Not sure what to do with them. Perhaps default text to them? I asked about it here: https://helsinkicity.slack.com/archives/C013682SYSD/p1616757237019800

nenonja (changes requested):

src/common/translation/i18n/en.json
nenonja: Funny enough the finnish and swedish translations do not have exclamation mark, so not sure about that either :)

melniiv: i was thinking it was a typo, as it was only in one language. i will confirm and fix in case

src/common/translation/i18n/fi.json
melniiv: in translations only 3 landing pages for /courses /events and /collections. but we have /home as well, so i kept the same translations as "default" for that.

src/domain/landingPage/landingPageMeta/LandingPageMeta.tsx
melniiv: they work differently, those you mentioned are coming from Wagtail. To make proper implementation (we discussed it with Khin and Niko) we need to make changes to the wagtail arch, the description was not supported for multiple Landing pages, but its supported for all other types. But its a bigger change, Mari suggested that we keep it easy, as the cms will be changed soon anyway.

melniiv:
duplicate

PR #252, opened 2021-03-29
TH-929: Fix meta descriptions

melniiv:

### DIFF
--- a/src/domain/app/layout/PageWrapper.tsx
+++ b/src/domain/app/layout/PageWrapper.tsx
@@ -12,12 +12,14 @@ import { useURIComponents } from '../../../hooks/useURIComponents';
 interface Props {
   className?: string;
   title?: string;
+  description?: string;
 }
 
 const PageWrapper: React.FC<Props> = ({
   children,
   className,
   title = 'appName',
+  description,
 }) => {
   const { t } = useTranslation();
   const location = useLocation();
@@ -50,8 +52,8 @@ const PageWrapper: React.FC<Props> = ({
       <Helmet>
         <html lang={locale} />
         <title>{translatedTitle}</title>
+        {description && <meta name="description" content={t(description)} />}
         <meta name="twitter:card" content="summary" />
-
         <link rel="canonical" href={host + url} />
         <link rel="alternate" hrefLang="fi" href={`${host}/fi/${path}`} />
         <link rel="alternate" hrefLang="sv" href={`${host}/sv/${path}`} />
--- a/src/domain/app/layout/PageWrapper.tsx
+++ b/src/domain/app/layout/PageWrapper.tsx
@@ -52,7 +52,7 @@ const PageWrapper: React.FC<Props> = ({
       <Helmet>
         <html lang={locale} />
         <title>{translatedTitle}</title>
-        {description && <meta name="description" content={t(description)} />}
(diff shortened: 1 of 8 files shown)


src/domain/app/layout/PageWrapper.tsx
```diff
@@ -50,8 +52,8 @@ const PageWrapper: React.FC<Props> = ({
       <Helmet>
         <html lang={locale} />
         <title>{translatedTitle}</title>
+        {description && <meta name="description" content={t(description)} />}
```
nenonja, line 55: Perhaps `meta.default.description` could be used as default, when description is not given as argument

src/domain/landingPage/landingPageMeta/LandingPageMeta.tsx
```diff
@@ -1,39 +1,57 @@
...
     landingPage.bottomBanner
   );
+
   const openGraphProperties = {
     description: metaInformation,
     image: [topBannerSoMeImage, bottomBannerSoMeImage],
     title: pageTitle,
   };
 
+  const getDescription = () => {
```
nenonja, line 42: This does not seem to work when I tried. 
Also I think that the idea of the `LandingPageMeta` is to read dynamically LandingPageQuery keywords to the meta-data, so perhaps you should use `<PageWrapper decription=...>` only for static metadata. Also I think it is much easier, you dont need any conditional logic with it.

nenonja (changes requested):
Some comments

src/domain/landingPage/landingPageMeta/__tests__/LandingPageMeta.test.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { LandingPageFieldsFragment } from '../../../../generated/graphql';
 import { fakeBanner, fakeLandingPage } from '../../../../test/mockDataUtils';
-import { actWait } from '../../../../test/testUtils';
+import { actWait, render } from '../../../../test/testUtils';
 import LandingPageMeta from '../LandingPageMeta';
 
 const landingPageTitle = 'Landing page title';
-const landingPageDescription = 'Landing page description';
+// this value is localized.
+const landingPageDescription =
```
nenonja, line 10: This test could be moved to PageWrapper test accordingly

src/domain/landingPage/landingPageMeta/LandingPageMeta.tsx
melniiv: just tested. it works. I don't think its a correct way to put the logic of the LandingPage to the PageContainer. This fix with local files is temporary, later it will come from new CMS. So it is still dynamic and still related only to landing pages.

src/domain/landingPage/landingPageMeta/__tests__/LandingPageMeta.test.tsx
melniiv: see the comment about getDescription()

src/domain/app/layout/PageWrapper.tsx
melniiv: fixed

Sartonon (approved):
Looks good. For me it is ok to go with this solution for now. Lets update it when we get metadata from cms.

PR #254, opened 2021-03-29
TH-1038: Hide map from internet events.

melniiv:

### DIFF
--- a/src/domain/event/eventContent/EventContent.tsx
+++ b/src/domain/event/eventContent/EventContent.tsx
@@ -22,6 +22,8 @@ const EventContent: React.FC<Props> = ({ event, eventType }) => {
   const locale = useLocale();
   const { description, photographerName } = getEventFields(event, locale);
 
+  const isInternetEvent = event?.location?.id === 'helsinki:internet';
+
   return (
     <div className={styles.eventContent}>
       <Container>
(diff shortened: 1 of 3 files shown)


src/domain/event/eventContent/EventContent.tsx
```diff
@@ -22,6 +22,8 @@ const EventContent: React.FC<Props> = ({ event, eventType }) => {
   const locale = useLocale();
   const { description, photographerName } = getEventFields(event, locale);
 
+  const isInternetEvent = event?.location?.id === 'helsinki:internet';
```
Sartonon, line 25: `'helsinki:internet'` could be stored as a constant, might be used somewhere else in the future.

Sartonon (commented):
Good. Maybe store 'helsinki:internet' to a constant.

src/domain/event/eventContent/EventContent.tsx
melniiv: fixed

Sartonon (approved):

src/domain/event/eventContent/__tests__/EventContent.test.tsx
```diff
@@ -88,3 +87,16 @@ it('should render event content fields', async () => {
     expect(screen.queryAllByRole(role, { name })).toHaveLength(2);
   });
 });
+
```
nenonja, line 90: we could also have another counterpart test that checks that map _is_ present when id is something else

nenonja (commented):
One comment

PR #257, opened 2021-03-30
Fix false positive footer & header tests

nenonja:
## Description :sparkles:

Some feature flag component tests give false positive result. These are now fixed.


Sartonon (approved):
Great :)

PR #259, opened 2021-03-31
TH-1062 | Add aluemeri to division blocklist

Sartonon:
## Description :sparkles:

- Delete Aluemeri from area dropdown
- Add test for the division hook

### DIFF
--- a/src/hooks/useDivisionOptions.tsx
+++ b/src/hooks/useDivisionOptions.tsx
@@ -1,6 +1,10 @@
-import { useNeighborhoodListQuery } from '../generated/graphql';
+import sortBy from 'lodash/sortBy';
+
+import { Neighborhood, useNeighborhoodListQuery } from '../generated/graphql';
 import getLocalisedString from '../util/getLocalisedString';
 import useLocale from './useLocale';
+
+const AREA_BLOCKLIST = ['kaupunginosa:aluemeri'];
 
 type DivisionOption = {
   text: string;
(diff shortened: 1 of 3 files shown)


src/hooks/useDivisionOptions.tsx
```diff
@@ -10,15 +14,20 @@ type DivisionOption = {
...
   const { data: neighborhoodsData } = useNeighborhoodListQuery();
+  const neighborhoodList = neighborhoodsData?.neighborhoodList.data;
+  const filteredNeighborhoodOptionList = neighborhoodList?.filter((option) =>
+    DIVISION_BLOCKLIST.every((blockOption) => blockOption !== option.id)
+  );
+
+  const neighborhoodToOptionValue = (neighborhood: Neighborhood) => ({
+    text: getLocalisedString(neighborhood.name, locale),
+    value: neighborhood.id,
+  });
```
nikomakela, line 25: I think it is needless here to create a new variable that is used only once with a map function.

nikomakela (approved):
Looks approvable.

src/hooks/useDivisionOptions.tsx
```diff
@@ -1,7 +1,11 @@
-import { useNeighborhoodListQuery } from '../generated/graphql';
+import sortBy from 'lodash/sortBy';
+
+import { Neighborhood, useNeighborhoodListQuery } from '../generated/graphql';
 import getLocalisedString from '../util/getLocalisedString';
 import useLocale from './useLocale';
 
+const DIVISION_BLOCKLIST = ['kaupunginosa:aluemeri'];
```
nikomakela, line 7: If there ever comes more of these, I would move it to some const-file.

nikomakela (approved):

src/hooks/useDivisionOptions.tsx
Sartonon: Purpose is to only make it more readable, but maybe going too far?

src/hooks/useDivisionOptions.tsx
Sartonon: Hmmm maybe, but I think this is only going to be used in this helper hook and it only has only this one const. So is there  benefit by putting it to another file? Or do you mean like some global const file?

src/hooks/useDivisionOptions.tsx
Sartonon: Updated :)

PR #261, opened 2021-03-31
TH-1056: Hide enrolment link from other info

melniiv:
## Description :sparkles:
The translations are added in case if, at some point, the link will be shown in other info, so the label of the link is correct.


src/domain/event/eventInfo/EventInfo.tsx
```diff
@@ -253,7 +255,8 @@ const EventInfo: React.FC<Props> = ({ event, eventType }) => {
             ))}
             {externalLinks.map((externalLink, index) => {
               return (
-                !!externalLink.link && (
+                !!externalLink.link &&
```
Sartonon, line 258: is `!!externalLink.link &&` redundant here?

melniiv: i am not sure can the link prop be empty or not.

Sartonon (approved):

PR #272, opened 2021-05-20
Fix apollo client loading bug (cache)

Sartonon:
## Description :sparkles:

- Fix apollo client loading bug (cache)


nikomakela (approved):
Yay!

PR #279, opened 2021-05-26
Hide the amount of max attendees and remaining spots from other info section.

nikomakela:
TH-1115.

NOTE: I think it would be good to resolve and merge LinkedEvents V2 changes before this PR is merged.

### DIFF
--- a/src/domain/event/eventInfo/EventInfo.tsx
+++ b/src/domain/event/eventInfo/EventInfo.tsx
@@ -12,7 +12,6 @@ import {
   IconTicket,
 } from 'hds-react';
 import { createEvent, EventAttributes } from 'ics';
-import isNil from 'lodash/isNil';
 import React from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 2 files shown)


nikomakela:
There are these unused fields in `QUERY_COURSE_DETAILS`, but I think they could be left there, since they are not very costly and it seems like they are wanted back when the stored data is filled with the information.

-       maximumAttendeeCapacity
-       minimumAttendeeCapacity
-       remainingAttendeeCapacity

```
export const QUERY_COURSE_DETAILS = gql`
  fragment courseFields on EventDetails {
    ...eventFields
    audienceMinAge
    audienceMaxAge
    extensionCourse {
      enrolmentStartTime
      enrolmentEndTime
      maximumAttendeeCapacity
      minimumAttendeeCapacity
      remainingAttendeeCapacity
    }
  }
  query CourseDetails($id: ID!, $include: [String]) {
    courseDetails(id: $id, include: $include) {
      ...courseFields
    }
  }
`;
```

Sartonon (approved):
Looks good!

PR #280, opened 2021-05-27
Added an util that resolves event type and can be used by collections EventList.

nikomakela:
TH-1138.

NOTE: The code does not work yet. It needs LinkedEvents v2 and event type work done before it can be continued. It also needs changes from proxy https://github.com/City-of-Helsinki/events-helsinki-api-proxy/pull/64 and cms City-of-Helsinki/events-helsinki-cms#23

### DIFF
--- a/src/domain/collection/eventList/EventList.tsx
+++ b/src/domain/collection/eventList/EventList.tsx
@@ -32,7 +32,7 @@ const EventList: React.FC<Props> = ({ collection }) => {
     collection,
     locale
   );
-  const eventType = getEventTypeFromRouteUrl(eventListQuery);
+  const eventType = getEventTypeFromRouteUrl(eventListQuery, locale);
   const searchParams = new URLSearchParams(
     eventListQuery ? new URL(eventListQuery).search : ''
   );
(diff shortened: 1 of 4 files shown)


src/domain/eventSearch/utils.tsx
```diff
@@ -443,3 +446,10 @@ export const getSearchQuery = (filters: Filters): string => {
 
   return buildQueryFromObject(newFilters);
 };
+
+export const getEventTypeFromRouteUrl = (url: string): EventType => {
+  if (matchPath(url, { path: ROUTES.COURSES, exact: true, strict: false })) {
+    return 'course';
+  }
+  return 'event';
+};
```
Sartonon, line 455: Maybe object map and looping could be used here. Would be easier to 'all' in the future. for example:

```
const routeToEventType = {
  [ROUTES.COURSES]: 'coure;
  [ROUTES.EVENTS]: 'event';
};
```

Sartonon (commented):
Just one suggestion for gwtting event type but not necessary

src/domain/eventSearch/utils.tsx
nikomakela: Added some changes.

nikomakela:
Here are the events listed in Kesätekemistä lapsille ja nuorille -collection: 
image

Here are the courses listed in Kesäleirit -(draft) collection: 
image

So the one solution now fits for both!

nikomakela:
The tests fails because new environment variables are missing from review, staging and production environments!

src/domain/collection/eventList/EventList.tsx
```diff
@@ -39,13 +44,13 @@ const EventList: React.FC<Props> = ({ collection }) => {
...
-      eventType: 'event',
+      eventType: eventType ?? 'event',
     });
-  }, [locale, searchParams]);
+  }, [eventType, locale, searchParams]);
 
   const { data: eventsData, fetchMore, loading } = useEventListQuery({
     notifyOnNetworkStatusChange: true,
-    skip: !eventListQuery,
+    skip: !eventListQuery || !eventType,
```
Sartonon, line 53: Maybe no need to invlude both  variables in this check?

Sartonon (approved):

src/domain/collection/eventList/EventList.tsx
nikomakela: It's there because eventType can be undefined when the url is given, but it's a bad one.

PR #284, opened 2021-05-31
Changed course page providers link text.

nikomakela:
TH-1140 To guide the user better, the course page publisher search link text is changed to something that describes that next query will be about courses.

NOTE: This is a self made not requested improvement.


Sartonon (approved):
GREAT!

PR #286, opened 2021-06-01
Update react scripts

Sartonon:
## Description :sparkles:

- Update react scripts package
- Update linter packages
- Clean up eslint config

## Issues :bug:
**[TH-1143](https://helsinkisolutionoffice.atlassian.net/browse/TH-1143):** 
**[TH-1141](https://helsinkisolutionoffice.atlassian.net/browse/TH-1141):** 


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-  "editor.formatOnSave": false,
+  "editor.formatOnSave": true,
```
nikomakela, line 2: my personal thought is that we should exclude editor settings from the repo. Not everybody uses vscode, so why should they have settings for it.

src/icons/IconMuseum.tsx
```diff
@@ -17,3 +17,5 @@ export default ({ className = '' }: Props): ReactElement => (
     </g>
   </svg>
 );
+
+export default IconMuseum;
```
nikomakela, line 21: export default change is fine for me, but I don't understand what's the difference between these 2.

nikomakela (approved):

.vscode/settings.json
Sartonon: Yeah but most use it and this one is pretty important configuration. No one is going to use vscoee without this config in this project.

Sartonon: But I could delete this and add some recommended stuff to README for vscode

src/icons/IconMuseum.tsx
Sartonon: Eslint started comlaining about these after updates and named default exports are better so they are not anonymous functions.

Sartonon: For example Nextjs complains about this by default

Sartonon: with `export default () => {}` function won't have any name in callstack when for example error happens and callstack is logged.

Sartonon: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-anonymous-default-export.md

.vscode/settings.json
Sartonon: Just checked that basically all city-of-helsinki projects have this directory included. It has all the basic setting to get vscode work nicely with our setup so I wouldn't remove it. All vscode users would need to add those settings anyway

PR #290, opened 2021-06-03
Links to super event and sub events

nikomakela:
TH-1029. TH-1147.

This pull requst also needs changes to the proxy server https://github.com/City-of-Helsinki/events-helsinki-api-proxy/pull/67


nikomakela:
Since LinkedEvents v2 is filtering events to with General type by default when an event type is not given as a filter parameter, it is needed that we change our EventType to the array type!

I will write tests later when the refactoring is done.

src/domain/event/EventPageContainer.tsx
```diff
@@ -65,9 +65,7 @@ const EventPageContainer: React.FC<EventPageContainerProps> = ({
   );
 
   React.useLayoutEffect(() => {
-    const isCoursePage = eventType === 'course';
-    // Only course page uses super event
-    if (superEventId && isCoursePage) {
```
nikomakela, line 70: Course has some superEvent features related to dates that general event should not have! I think it should be left so, so this still needs some changes. It would be just nice to roll the super event once queried here to a subcomponent.

src/domain/event/EventUtils.ts
```diff
@@ -403,3 +409,10 @@ export const getAudienceAgeText = (
   }`;
   return `${ageLimit} -${t('event.info.age')}`;
 };
+
+export const getEventTypeByEventTypeId = (
+  eventTypeId: EventTypeId
+): EventType =>
+  Object.keys(EVENT_TYPE_TO_ID).find(
+    (eventType) => EVENT_TYPE_TO_ID[eventType as EventType] === eventTypeId
+  ) as EventType;
```
Sartonon, line 418: Hmm I don't understand why this is needed. Is it to exclue 'All' type or something?

src/domain/event/queryUtils.ts
```diff
@@ -72,46 +72,63 @@ export const useSimilarEventsQuery = (
...
+      // when no event type is given as a parameter,
+      // this needs a list of eventTypes
+      // or otherwise some events or courses are always excluded.
+      // eventType: EventTypeId.Course //event.typeId
+    }),
+    [event.id]
+  );
+
+  return { superEventId: event.id, variables };
+};
```
Sartonon, line 118: I'm not sure if the is need for this helper function

nikomakela: No there is no need for it. I just "started from somewhere" and made it equal with the old implementation of "other times of the event". I'll remove those both. 👍

nikomakela: Actually, after rethinking this, I think I keep them both there, because I really need `useOtherEventTimesVariables` and it has several dependencies. I would not like to import those dependencies in subEvents and like this its easy to see what is the difference between those 2. I can move them if you don't agree

src/domain/event/EventUtils.ts
nikomakela: Event object that is typed with EventFields now has a field called typeId, which is 'General' or 'Course'. In some cases, e.g when I'm fetching a list of other event times, some of the events might have a different type than the related event. Instead of hard coding and typing the EventType for the event, I would like to use the one stored in database and there I need that `getEventTypeByEventTypeId` -function, which converts 'General' to 'event' and ''Course' to a 'course'. Like that you can have a general event that has some related courses and when the course is clicked, user will be taken to a course section instead of event section - it will change the url to the right one!

src/domain/event/EventPageContainer.tsx
nikomakela: Moved this if-statement to EventHero.

nikomakela:
Super event and sub events both have their own section and they are sharing the event list styles with other event times:

image

No matter what event type the super event or the sub events have, they are always taken to right url path.

nikomakela:
I'm going to add the missing tests if the approach is approved.

src/domain/event/EventUtils.ts
Sartonon: Oh yeah I was little bit blind here, didn't notice that this returns different type than takes as a parameter

src/domain/event/eventInfo/EventInfo.tsx
```diff
@@ -52,39 +60,60 @@ const EventInfo: React.FC<Props> = ({ event, eventType }) => {
...
-  };
+  return (
+    <div className={styles.eventInfo} ref={eventInfoContainer}>
+      <div className={styles.contentWrapper}>
+        <DateInfo event={event} eventType={eventType} />
+        <SuperEvent superEvent={superEvent} />
+        <SubEvents event={event} />
+        <OtherEventTimes event={event} />
+        {(audienceMinAge || audienceMaxAge) && <AudienceAgeLimitations audienceMinAge={audienceMinAge} audienceMaxAge={audienceMaxAge} eventType={eventType} />}
+        <LocationInfo event={event}/>
```
Sartonon, line 85: Prettie not used?

src/domain/event/queryUtils.ts
Sartonon: Thinking if React.useMemo is at all useful here. `useOtherEventTimes` could do without both useMemos and make code easier. I don't think there is any noticeable performance difference

src/domain/event/eventInfo/EventsHierarchy.tsx
```diff
@@ -0,0 +1,119 @@
...
+export const superEventTestId = 'super-event';
+
+const SubEvents: React.FC<{ event: EventFields }> = ({ event }) => {
+  const { t } = useTranslation();
+  const [isListOpen, setIsListOpen] = React.useState(false);
+  const eventType = event.typeId
+    ? getEventTypeByEventTypeId(event.typeId)
+    : 'event';
+
+  const { superEventId, variables } = useSubEventsQueryVariables(event);
```
Sartonon, line 26: hmm could be moved inside useSubEvents instead of calling it here and passing many params?

src/domain/event/queryUtils.ts
```diff
@@ -72,46 +72,67 @@ export const useSimilarEventsQuery = (
...
-    [eventType, superEventId]
+    [superEventId]
   );
 
   return { superEventId, variables };
 };
 
-// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
-export const useOtherEventTimes = (
+export const useSubEventsQueryVariables = (
```
Sartonon, line 101: could be simplified:

```
export const useSubEventsQueryVariables = (
  event: EventFields
): { superEventId: string | undefined; variables: EventListQueryVariables } => {
  const variables = {
    sort: EVENT_SORT_OPTIONS.START_TIME,
    start: 'now',
    superEvent: event.id,
    /* 
      Since LE v2 is listing general type of events 
      when no event type is given as a parameter,
      this needs a list of eventTypes
      or otherwise some events or courses are always excluded.
      */
    eventType: [EventTypeId.General, EventTypeId.Course], //event.typeId
  };

  return { superEventId: event.id, variables };
};
```

src/domain/event/queryUtils.ts
```diff
@@ -72,46 +72,67 @@ export const useSimilarEventsQuery = (
   return { data, loading };
 };
 
-const useOtherEventTimesVariables = (
-  event: EventFields,
-  eventType: EventTypeId = EventTypeId.General
-) => {
+const useOtherEventTimesVariables = (event: EventFields) => {
```
Sartonon, line 75: could be simplified:

```
const useOtherEventTimesVariables = (event: EventFields) => {
  const superEventId = getEventIdFromUrl(
    event.superEvent?.internalId || '',
    'event'
  );

  const variables = {
    sort: EVENT_SORT_OPTIONS.START_TIME,
    start: 'now',
    superEvent: superEventId,
    /* 
    Since LE v2 is listing general type of events 
    when no event type is given as a parameter,
    this needs a list of eventTypes
    or otherwise some events or courses are always excluded.
    */
    eventType: [EventTypeId.General, EventTypeId.Course], //event.typeId
  };

  return { superEventId, variables };
};
```

Sartonon: No need for useMemos

src/domain/event/eventInfo/EventsHierarchy.tsx
Sartonon: hmm never mind, it is used in 2 different places.

src/domain/event/eventInfo/EventsHierarchy.tsx
```diff
@@ -0,0 +1,119 @@
...
+    </div>
+  );
+};
+
+const SuperEvent: React.FC<{ superEvent: SuperEventResponse | undefined }> = ({
+  superEvent,
+}) => {
+  const { t } = useTranslation();
+
+  if (!superEvent || !superEvent.data) return null;
```
Sartonon, line 95: ` if (!superEvent?.data) return null;`
Would be enough?

src/domain/event/eventInfo/eventList/EventList.tsx
```diff
@@ -0,0 +1,102 @@
...
+              locale,
+              timeAbbreviation: t('commons.timeAbbreviation'),
+            })
+          : '';
+        return (
+          <li key={event.id}>
+            <button
+              className={styles.listButton}
+              // Events can have different event type id than it's super event or sibling has - purposely or accidentally.
+              onClick={() => moveToEventPage(event.id, event?.typeId ? getEventTypeByEventTypeId(event.typeId) : 'event')}
```
Sartonon, line 87: prettier?

src/domain/event/eventInfo/EventInfo.tsx
nikomakela: I had some unexpected problems with editor's auto formatter and it is committed with no-verify. Need to run prettier everywhere when committing tests.

src/domain/event/eventInfo/eventList/EventList.tsx
nikomakela: Yep, need to run it.

src/domain/event/eventInfo/EventInfo.tsx
nikomakela: Run prettier to all files. Should be ok now.

src/domain/event/eventInfo/eventList/EventList.tsx
nikomakela: Ran prettier to all files. Should be ok now.

nikomakela:
Old tests are now (finally) fixed. Now I start writing the new ones.

nikomakela:
One more reminder: This needs a new version from proxy in order to work.

src/domain/event/eventInfo/__tests__/EventInfo.test.tsx
```diff
@@ -364,3 +391,225 @@ describe('OrganizationInfo', () => {
...
+      )
+    );
+    expect(history.location.pathname).toBe(`/fi/events/${superEvent.id}`);
+  });
+
+  it('should should not render super event title when super event is not given', async () => {
+    render(<EventInfo event={event} eventType="event" />, {
+      mocks,
+    });
+    await actWait();
```
Sartonon, line 433: Nothing asyncronous happening in this case? Or is it added to be futureproof somehow? :)

src/domain/event/eventInfo/__tests__/EventInfo.test.tsx
```diff
@@ -364,3 +391,225 @@ describe('OrganizationInfo', () => {
...
+  it('should render sub events title and content when sub events are given', async () => {
+    render(<EventInfo event={event} eventType="event" />, {
+      mocks: mocksWithSubEvents,
+    });
+    await actWait();
+    expect(
+      screen.queryByRole('heading', {
+        name: translations.event.subEvents.title,
+      })
+    ).toBeInTheDocument();
```
Sartonon, line 520: `await findBy` could be used

src/domain/event/eventInfo/__tests__/EventInfo.test.tsx
```diff
@@ -364,3 +391,225 @@ describe('OrganizationInfo', () => {
...
+  ])(
+    'should navigate to sub events page when it is clicked',
+    async (eventTypeId: EventTypeId, url: string) => {
+      const { history } = render(
+        <EventInfo event={event} eventType="event" />,
+        {
+          mocks: mocksWithSubEvents,
+        }
+      );
+      await actWait();
```
Sartonon, line 536: Here we could replace actWait with findBy. I think it is clearer what we are waiting :)
`const eventsList = await screen.findByTestId(subEventsListTestId);`

Sartonon (commented):
Looking good. Some comments/suggestions added, mostly related to testing.

src/domain/event/EventUtils.ts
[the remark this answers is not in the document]
nikomakela: I tried to avoid that the change is needed in multiple places. I understand that it's hard to understand what it does, but the thing is that you should never touch it :) @Sartonon, I can change it to your suggestion if you really wish so.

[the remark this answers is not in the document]
nikomakela: Added a function description.

src/domain/event/eventInfo/__tests__/EventInfo.test.tsx
```diff
@@ -364,3 +391,225 @@ describe('OrganizationInfo', () => {
...
+      10,
+      range(1, 11).map((i) => ({
+        endTime: addDays(new Date(endTime), i).toISOString(),
+        startTime: addDays(new Date(startTime), i).toISOString(),
+        typeId: i % 2 === 0 ? EventTypeId.Course : EventTypeId.General,
+        superEvent: { internalId: superEventInternalId },
+      }))
+    ),
+    meta,
+  };
```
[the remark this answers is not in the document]
nikomakela: Moved them to utils -folder. Couldn't use them from __mocks__ -forlder because of this : "imported from a __mocks__ directory. Instead use `jest.mock` and import from the original module path jest/no-mocks-import"

src/domain/event/EventUtils.ts
```diff
@@ -410,6 +410,16 @@ export const getAudienceAgeText = (
...
+/**
+ * EVENT_TYPE_TO_ID configuration object is a mapper between event type and eventTypeId
+ * where an EventType instance is a key and EventTypeId instance is a value.
+ * To avoid writing yet another configuration table, this function does the opposite -
+ * it flips the configuration around and the value can be used to find the related key.
+ *
+ * E.g. The event objects contains a typeId that is an instance of EventTypeId,
+ * so it is sometimes needed to get a related event type for an eventTypeId instance
+ * and this function does exactly that.
+ */
```
Sartonon, line 422: Could also have example:

```
EventTypeId -> EventType
'General'       -> 'event'
```

PR #295, opened 2021-06-06
TH-1061 | Add remote events checkbox to events search

Sartonon:
## Description :sparkles:

- Add remote events checkbox to events search

## Issues :bug:

**[TH-1061](https://helsinkisolutionoffice.atlassian.net/browse/TH-1061):** 

## Screenshots :camera_flash:

Screenshot 2021-06-06 at 17 12 38


nikomakela (commented):
The code changes looks good, but since the internet events are in their own cache and so called locale events are in their own, I need to test this a bit. This implementation seems to mix those 2. I wonder if we should use a query meant for "Ongoing internet events".

PR #309, opened 2021-06-17
Alternative implemenation for landing page categories

nikomakela:
TH-1151 Alternative implemenation for landing page categories by using centralized configs and objects from utils.


nikomakela:
I'm not happy about the fact that hobbyTypes are still in their own group, but it's almost mandatory, because they are wanted to be listed in a different filter input field.

Sartonon (approved):
Much better :)

PR #315, opened 2021-06-21
Accessibility audition fixes

nikomakela:
TH-1004: Fixes based on UX audit
TH-1167: aria-label for Helsinki logo.
TH-1168: aria-label for language selector.
TH-1169 aria-label for Beta-button
TH-1170 Added area-expanded for every search criteria.
TH-1171 Removed aria-expanded from popular categories (mobile view).
TH-1172 aria-hidden attribute for super event dates in single hobby page.


Sartonon (approved):
Looking good!

PR #316, opened 2021-06-22
TH-1174 | Update hds and use links where necessary

Sartonon:
## Description :sparkles:

- Update hds and use links where necessary
- Some refactoring


nikomakela (commented):
The changes are looking good, but github cannot build.

PR #319, opened 2021-06-30
Fixed TS compilation error on browser tests.

nikomakela:
TH-1103. There was a circular import, which wrecked the css import. Also the parameter type was wrong in normalizeSuitableFor - it was a list of string or numbers, but it should be a list of mixed items..


Sartonon (approved):
AAmaazing :D

PR #321, opened 2021-07-16
disable autocomplete from datepickers to make screen reader work better

Sartonon:
### Closes :no_good_woman:

**[DEV-XXX](https://helsinkisolutionoffice.atlassian.net/browse/DEV-XXX):**


nikomakela (approved):
Yep, now it really does work better! Fixed my issues with browser's autocomplete!

PR #323, opened 2021-12-28
fix: show only the upcoming events in a list of curated events

nikomakela:
TH-1183. Added a start parameter to events by ids query.

NOTE: Needs changes from proxy https://github.com/City-of-Helsinki/events-helsinki-api-proxy/pull/82


Sartonon (approved):
Great!

PR #326, opened 2022-01-13
fix(header): updated the header links

nikomakela:
TH-1192. Added the app name and the about link to the header. Changed the events link text.


Sartonon (approved):
Looks good 👍

PR #338, opened 2022-03-28
TH-1199 | Replace self hosted Matomo with Digia Iiris Matomo

Jokinen:
## Description :sparkles:

Changes from self hosted Matomo into Digia Iiris Matomo.

I tested this configuration locally. Aiski verified that Matomo received data with these configs.

Contains documentation for release 0.10.0.

### Closes :no_good_woman:

**[TH-1199](https://helsinkisolutionoffice.atlassian.net/browse/TH-1199):**


nikomakela (approved):
Looking good

